PR #3, opened 2015-01-27
[CORE] Add getRow helper to GenerateFixedMatrixNxN

nknize:
Add boilerplate to GenerateFixedMatrixNxN to generate companion 'getRow' helper method


lessthanoptimal:
I've added the functionality to FixedOps instead and reworked it.  This implementation broke the procedural interfaces goal of (almost) never forcing the user to declare new memory.

PR #12, opened 2016-02-11
update gradle to show full example

ersin-ertan:
Example shows the group, artifactId, and version, consolidated into one line.
Useful for people new to the gradle build system.

Website would need to reflect the changes for the example as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,10 @@ Gradle:
 ```
 compile group: 'org.ejml', name: 'all', version: '0.29'
 ```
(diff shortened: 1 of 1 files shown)


lessthanoptimal:
Hmm the original gradle code was actually legal.  There are couple of ways you can specify dependencies and either notations should work just fine.

ersin-ertan:
Learned something new. Thanks, I have only seen the condensed version, thus it might help someone out that makes the same mistake.

PR #47, opened 2018-10-15
fixed bug in inversion of 1D covariance matrices

ciamac:
Hi,
There is a minor bug when inverting a 1D covariance matrix due to a typo, this patch should fix it.

Best,
Ciamac.


lessthanoptimal:
thanks!

PR #52, opened 2019-01-17
Add abs() operation for DMatrixRMaj in CommonsOps_DDRM

vfrico:
Add two variants for the absolute value of a matrix. The result is the absolute value for each element.


lessthanoptimal:
thanks! sorry it took a bit to merge in

PR #56, opened 2019-06-17
Fix exception content

scharron:
The exception content is wrong, resulting in interesting stuff like : `Unexpected dimensions for X: X rows = 2 expected = 2`


lessthanoptimal:
thanks for the fix!

PR #71, opened 2020-07-06
Enable tests running via gradle

FlorentinD:
This change makes it possible to run the test via "./gradlew test".
Working in Intelij, tests are executed by default via gradle.
Previously no tests could be found for this project.

In addition, this change allows running the test in CI.

Currently `org.ejml.data.TestDMatrixSparseCSC#growMaxLength_veryLarge` throws an OutOfMemory Exception.
As a `double` needs 8 bytes and the matrix in the test has the dimensions `10^9 x 10^9`, it might be better to just disable the test.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,9 @@ subprojects {
     test {
         ignoreFailures true
         reports.html.enabled = false
+        useJUnitPlatform {
+            includeEngines 'junit-jupiter'
+        }
     }
 
     repositories {


lessthanoptimal:
Merging! I always tell intellij to not use Gradle since it's much faster if I don't.

I'm not getting an out of memory error for `TestDMatrixSparseCSC#growMaxLength_veryLarge` but I might just have more memory in my machine. Just did some quick math and the column indexes would take up about 4G of RAM. The double data array is negligible since it's length is 4 initially.  I'm reducing the number of columns by a factor of 10 to make it a more reasonable test on machines with less memory.

FlorentinD:
Thanks for the tip of changing the test execution to use intellij.

I misread the test, as I thought `growMaxLength` would let the data array grow to its dimension (which also doesn't really make sense for sparse matrices).
Reducing the number of columns by a factor of 10 resolved the problem :+1: 
My RAM should have been sufficient though, but I think gradle uses a default heap size of 1GB.

PR #74, opened 2020-07-27
Allow a fallbackValue for `get()` on sparse matrices

FlorentinD:
Sometimes, an unassigned entry does not mean `0` but f.i. `Double.MAX_VALUE` for `min()`.

I duplicated the existing `get(x,y)` and `unsafe_get()` methods.
An alternative would be, for the existing ones to call `get(x,y, 0)`, but I thougth to avoid additional function calls.
Better be cautious how good the compiler can inline the calls, right?

### DIFF
--- a/main/ejml-core/src/org/ejml/data/DMatrixSparse.java
+++ b/main/ejml-core/src/org/ejml/data/DMatrixSparse.java
@@ -28,6 +28,30 @@ import java.util.Iterator;
 public interface DMatrixSparse extends DMatrix, MatrixSparse {
 
     /**
+     * Returns the value of value of the specified matrix element.
+     *
+     * @param row           Matrix element's row index..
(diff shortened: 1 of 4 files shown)


lessthanoptimal (changes requested):
I'm approving the concept, but there should be unit tests and this should be added to `DMatrixSparse` which means it will also be added to other sparse matrices. Shouldn't be too bad of a change but I can do that if that's more work than you bargained for with this PR.

FlorentinD:
Thank you for your review.
It's no problem to make these changes myself :)

FlorentinD:
@lessthanoptimal I addressed your comment and squashed the commit.
Nice, that I could just add one test-case to `GenericTestsDMatrixSparse`.

lessthanoptimal:
thanks!

Issue #68, opened 2020-04-09
Support for IMatrix* and Graph Theory Operations


breandan:
I see that ejml has support for [logical matrices](https://github.com/lessthanoptimal/ejml/blob/SNAPSHOT/main/ejml-core/src/org/ejml/data/BMatrixRMaj.java). Would you be open to including integer matrices, or is this currently out of scope? These have a number of applications in graph theory and discrete math.

lessthanoptimal:
The support is fairly minimal, mostly as a way to mask elements in a matrix. However is there much overlap between linear algebra and the techniques you mentioned?

breandan:
There are a number of interesting connections between linear algebra and graph theory, e.g. you can take [eigenvalues of graph laplacians](http://www.math.ucsd.edu/~fan/research/cb/ch1.pdf) or [invert the adjacency matrix](https://core.ac.uk/download/pdf/153558187.pdf). For a short survey of some applications, see [Johnson](http://www.math.utah.edu/~gustafso/s2017/2270/projects-2017/dylanJohnson/Dylan%20Johnson%20Graph%20Theory%20and%20Linear%20Algebra.pdf) or maybe [Spielman](https://www.cs.yale.edu/homes/spielman/sagt/sagt.pdf) if you're looking for something more comprehensive.

edit: You could probably use `FMatrix*` or `DMatrix*`, but I'm not sure if certain operations might be numerically less stable than having a native integer matrix type. Otherwise I guess it shouldn't be a major issue.

breandan:
You might also check out the [GraphBLAS](http://graphblas.org/) library and the [Graph Algorithms in the Language of Linear Algebra](https://epubs.siam.org/doi/book/10.1137/1.9780898719918) book. There is a lot of interest in sparse matrix representations for graphs, it would be great to have a native linear algebra library with support for boolean and integer matrices on the JVM.

lessthanoptimal:
Do you @szarnyasg ? He's also interested in linear algebra and graphs. I'm not opposed to adding an integer matrix type, but all the algorithms to make it useful. Right now I just don't have a personal need.

szarnyasg:
Hi,

> it would be great to have a native linear algebra library with support for boolean and integer matrices on the JVM.

I agree but its a big undertaking. Adding an integer matrix type sounds fairly simple but then you also need to support add various for semirings and, if you want to apply the tool to solve big graph problems, it needs to run in parallel as well.

[...]

PR #75, opened 2020-07-28
Matrix operations with semi rings

FlorentinD:
Allowing to redefine + and * for operations over matrices, e.g. multiplication, elementwise-mult, add, ... .

The operations are based on the existing matrix operations by copying them and adding semi-rings.
The semi rings are added by replacing hard-coding + and * in the functions, but calling operators like in `apply` or `reduce`.

I also started to benchmark the operations (results at https://github.com/FlorentinD/ejml/blob/benchmarks/benchmark/results.txt).
It looks like they add no overhead, besides the obvious impact of the different + and * itself.
For instance `min, max` is unsuprisingly slower than `or, and`.
However these are only first benchmarks and also have an unusual high error rate, which I hope to be able to resolve.

With semi-rings, it is possible to add a cool graph example.
Basically the current tests are even based on examples, found here https://fosdem.org/2020/schedule/event/graphblas/attachments/slides/4132/export/events/attachments/graphblas/slides/4132/graphblas_introduction.pdf (f.i. slide 28).

The testing of the introduced methods is improve-able.


lessthanoptimal:
@szarnyasg and @breandan could you provide some feedback about this semiring implementation? I don't know enough about this domain. I'll be primarily looking at the PR from a coding standards perspective.

FlorentinD:
Basically the semi-ring is just a wrapper around two `DBinaryOperator`.

The slides I referenced in the description are even from @szarnyasg :smile:

lessthanoptimal:
yeah after I posted that message I saw that! Was confused for a second and thought you might be part of the same group then saw the schools were different.

FlorentinD:
No, the FOSDEM talk was just really good and mainly the reason, why I am looking into GraphBLAS for Java.

szarnyasg:
This is truly fantastic. I'll set aside some time during the weekend to review this PR.

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class TestMatrixMatrixMultWithSemiRing_DSCC {
```
breandan, line 38: Where is the matrix-matrix multiplication in this test? I would have expected this to test sparse-dense, dense-sparse, and dense-dense matmuls with semi-ring algebras.

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,82 @@
...
+        inputMatrix.set(5, 2, 0.5);
+        inputMatrix.set(6, 2, 1);
+        inputMatrix.set(6, 3, 1);
+        inputMatrix.set(6, 4, 1);
+
+    }
+
+    @ParameterizedTest(name = "{0}")
+    @MethodSource("sparseVectorMatrixMultSources")
+    public void mult_v_A(String desc, DSemiRing semiRing, double[] expected) {
```
breandan: This test looks like it belongs in `TestMatrixVectorMultWithSemiRing_DSCC`.

breandan (changes requested):
@FlorentinD Looks like the matrix-matrix tests are incomplete?

main/ejml-core/src/org/ejml/ops/DMonoids.java
```diff
@@ -0,0 +1,37 @@
...
+public final class DMonoids {
+    public static final DMonoid AND = new DMonoid(1, (a, b) -> (a == 0 || b == 0) ? 0 : 1);
+    public static final DMonoid OR = new DMonoid(0, (a, b) -> (a != 0 || b != 0) ? 1 : 0);
+    public static final DMonoid XOR = new DMonoid(0, (a, b) -> ((a == 0 && b == 0) || (a != 0 && b != 0)) ? 0 : 1);
+    public static final DMonoid XNOR = new DMonoid(0, (a, b) -> ((a == 0 && b == 0) || (a != 0 && b != 0)) ? 1 : 0);
+
+    public static final DMonoid PLUS = new DMonoid(0, Double::sum);
+    public static final DMonoid TIMES = new DMonoid(1, (a, b) -> a * b);
+
+    // TODO: performance incr. worth not using safe Math.min/max?
```
breandan: This depends on the JVM and should probably be benchmarked. Also, the semantics are different. 

cf. OpenJDK implementations below:

https://github.com/openjdk/jdk/blob/93904460818dfbd6bf028b06777238603ce78614/src/java.base/share/classes/java/lang/Math.java#L1569-L1580
https://github.com/openjdk/jdk/blob/93904460818dfbd6bf028b06777238603ce78614/src/java.base/share/classes/java/lang/Math.java#L1652-L1663

cc: @lessthanoptimal

main/autocode/src/org/ejml/GenerateJavaCode32.java
```diff
@@ -54,6 +54,14 @@ public GenerateJavaCode32() {
...
         prefix64.add("DBinary");
         prefix32.add("FBinary");
+        prefix64.add("DMonoid");
+        prefix32.add("FMonoid");
+        prefix64.add("DSemiRing");
+        prefix32.add("FSemiRing");
+        prefix64.add("DMonoids");
+        prefix32.add("FMonoids");
+        prefix64.add("DSemiRings");
+        prefix32.add("FSemiRings");
```
breandan, line 64: Just curious if you are planning to add `int` and `boolean` domains to this PR?

main/ejml-dsparse/src/org/ejml/sparse/csc/misc/ImplCommonOpsWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,183 @@
...
+import javax.annotation.Nullable;
+import java.util.Arrays;
+
+import static org.ejml.UtilEjml.adjust;
+import static org.ejml.sparse.csc.mult.ImplSparseSparseMultWithSemiRing_DSCC.multAddColA;
+
+/**
+ * based on ImplCommonOps_DSCC
+ */
+public class ImplCommonOpsWithSemiRing_DSCC {
```
breandan, line 35: Maybe this class should extend `ImplCommonOps_DSCC`?

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
FlorentinD: I created the sparse vector as a matrix, so I somewhat test sparse-sparse matmuls.
Dense-dense is currently not supported, as I only introduced semi-rings for sparse matrices yet.

Sparse-dense and dense-sparse sparse tests could be added of course.
In general it would be nice to have extensive tests for these functions, but I think won't have the time to do so currently.
Would be very glad for any help there!

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
FlorentinD: The describtion suggest so, but as I use a sparse matrix to represent the vector instead of primitive arrays, it think its valid to put it in here

main/ejml-core/src/org/ejml/ops/DMonoids.java
FlorentinD: I meant to remove the `TODO`, as I don't think I will address it. 
It was more a general comment, as I also saw that @lessthanoptimal used hard-coded min/max versions at other places.
As the library tries to avoid using unneccessary wrappers, I think using hard-coded versions is the "safe" way to go. 

I would be interested in results of those benchmarks nonetheless (probably already exist somewhere)

Thank you for pointing out, that the semantics actually differ!

main/autocode/src/org/ejml/GenerateJavaCode32.java
FlorentinD: No not in this PR, I think this should belong in a seperate PR.
However currently I don't plan to add more types of matrices, except for sparse boolean masks in another PR (implictly true for every defined cell).

I see the main problem with different types in having to support variants of all operators to support all combinations of types.
Guess this could also be handled by code generation, but requiring significant amout of work in my opinion.

main/ejml-dsparse/src/org/ejml/sparse/csc/misc/ImplCommonOpsWithSemiRing_DSCC.java
FlorentinD: What would be the value here?
I don't see how I can override any methods, as the signature differs regarding the semi-ring parameter.
I could overwrite the ones without semi-rings to use call the semi-ring version and call with the natural plus-times semi-ring

FlorentinD:
@breandan Thank you for taking the time to review my PR.
The testing in general is incomplete for this contribution, I hope this can be added later on.
My existing test can more be viewed as PoC of the functions.
Ideally the added operators should be more than equally tested as the existing operators.
I would appreciate any help on that part!

main/ejml-dsparse/src/org/ejml/sparse/csc/misc/ImplCommonOpsWithSemiRing_DSCC.java
breandan: I guess you could avoid the call to [tranpose](https://github.com/lessthanoptimal/ejml/blob/60093bc94a6b9577f08e69ef36371eec2ba4383d/main/ejml-dsparse/src/org/ejml/sparse/csc/misc/ImplCommonOps_DSCC.java#L47) since the signatures match, but probably it's not worth it. In general it's probably best to keep everything flat and use static methods to avoid confusion.

main/autocode/src/org/ejml/GenerateJavaCode32.java
breandan: There might be a way to avoid unnecessary duplication by using generics, but it would require a different API and would probably be more difficult to optimize on the JVM. Agree it makes sense to have a separate PR for #68.

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
breandan: Makes sense. Eventually it would be nice to have dedicated methods for handling dense-dense and dense-sparse/sparse-dense matmuls. But even though DSCC is not optimized for dense matrices, we would still expect a sparse matrix filled with nonzeroes to produce a numerically correct result.

main/autocode/src/org/ejml/GenerateJavaCode32.java
FlorentinD: The problem I know of, is that generics don't support primitive types.
I think project valhalla solves this problem, so maybe in a future JDK it will be possible. 
Also the return values from f.i. get() couldn't be primitives then.

main/ejml-dsparse/src/org/ejml/sparse/csc/misc/ImplCommonOpsWithSemiRing_DSCC.java
FlorentinD: I initially added the transpose method, as I wanted a common place for Graphblas operators.
However I also removed apply and reduce, as I didn't think having a redirect here is worth it.

I will remove the transpose function from this class, thank you for spotting this  :+1:

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
FlorentinD: @breandan I added some test, also including other operations and dense/sparse version of DSCC.
I came up with an easy way, by just using the PLUS-TIMES semi-ring and asserting against the existing implementation.
I think I can conclude the correctness for other semi-rings this way, without asserting on concrete results.

main/ejml-core/src/org/ejml/ops/DMonoids.java
breandan: Based on my tests (OpenJDK 11.0.7 amd64), `Math.max(a, b)` is quite a bit faster than the hard-coded version. @lessthanoptimal might want to look into that. There's some conflicting reports: https://stackoverflow.com/questions/22752198/java-math-min-max-performance

breandan:
Looks good to me! The only other thing I can think of is maybe going fully branchless (if possible). I noticed there are several conditionals in the multiplication inner loops. Not an expert but I've read this can have a significant performance impact.

lessthanoptimal:
When I'm back from vacation this week I'll go over the PR and everyone's comments.

@szarnyasg Any comments?

FlorentinD:
A deviation from the GraphBLAS C-API is that I used a monoid instead of an binary operator for defining the `mult` operator in a semi-ring. It does not seem to me like an issue though.

szarnyasg:
@FlorentinD on some occasions, the LAGraph library uses a binary operator for multiplication that is technically not a monoid. For example, BFS uses the "any" function, `any(x,y) = either x or y` as part of the "any-pair" semiring. The [SuiteSparse specification](http://mit.bme.hu/~szarnyas/grb/GraphBLAS_UserGuide.pdf#page=54) discusses this in detail:

> The result of the ANY monoid is non-deterministic, unless it is coupled with
> the PAIR multiplicative operator. In this case, the ANY_PAIR semiring will
> return a deterministic result, since f(1, 1) is always 1, for the ANY operator
> f(x, y).

That might not be a problem after all but it is something to keep in mind.

FlorentinD:
@szarnyasg I also couldn’t find any usage for the neutral element of the multiplication.
But using just a filler element for non real monoids, seemed viable to me.
Might be still worth to document in the semi-ring class.

main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOpsWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,262 @@
...
+    public static DMatrixSparseCSC mult(DMatrixSparseCSC A, DMatrixSparseCSC B, @Nullable DMatrixSparseCSC output, DSemiRing semiRing) {
+        return mult(A, B, output, semiRing, null, null);
+    }
+
+    /**
+     * Performs matrix multiplication.  output = A*B
+     *
+     * @param A        (Input) Matrix. Not modified.
+     * @param B        (Input) Matrix. Not modified.
+     * @param output   (Output) Storage for results.  Data length is increased if increased if insufficient.
```
szarnyasg: Typo: `is increased if increased`

main/ejml-core/src/org/ejml/ops/DMonoid.java
```diff
@@ -0,0 +1,34 @@
...
+
+    DMonoid(double id, DBinaryOperator func) {
+        this.id = id;
+        this.func = func;
+    }
+
+    DMonoid(DBinaryOperator func) {
+        this(0, func);
+    }
+}
```
szarnyasg: There's a missing newline at the end of file in many Java files.

main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOpsWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,262 @@
...
+     * @param semiRing Semi-Ring to define + and *
+     * @param gw       (Optional) Storage for internal workspace.  Can be null.
+     * @param gx       (Optional) Storage for internal workspace.  Can be null.
+     */
+    public static DMatrixSparseCSC mult(DMatrixSparseCSC A, DMatrixSparseCSC B, @Nullable DMatrixSparseCSC output, DSemiRing semiRing,
+                            @Nullable IGrowArray gw, @Nullable DGrowArray gx) {
+        if (A.numCols != B.numRows)
+            throw new MatrixDimensionException("Inconsistent matrix shapes. " + stringShapes(A, B));
+        output = reshapeOrDeclare(output,A,A.numRows,B.numCols);
```
szarnyasg: missing spaces

main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOpsWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,262 @@
...
+        ImplSparseSparseMultWithSemiRing_DSCC.mult(A, B, output, semiRing, gw, gx);
+
+        return output;
+    }
+
+    public static DMatrixSparseCSC multTransA(DMatrixSparseCSC A, DMatrixSparseCSC B, DMatrixSparseCSC output, DSemiRing semiRing,
+                                     @Nullable IGrowArray gw, @Nullable DGrowArray gx) {
+        if (A.numRows != B.numRows)
+            throw new MatrixDimensionException("Inconsistent matrix shapes. " + stringShapes(A, B));
+        output = reshapeOrDeclare(output,A,A.numCols,B.numCols);
```
szarnyasg: missing spaces

main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOpsWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,262 @@
...
+     * @param A        Matrix
+     * @param B        Dense Matrix
+     * @param semiRing Semi-Ring to define + and *
+     * @param output   Dense Matrix
+     */
+    public static DMatrixRMaj mult(DMatrixSparseCSC A, DMatrixRMaj B, @Nullable DMatrixRMaj output, DSemiRing semiRing) {
+        if (A.numCols != B.numRows)
+            throw new MatrixDimensionException("Inconsistent matrix shapes. " + stringShapes(A, B));
+
+        output = reshapeOrDeclare(output,A.numRows,B.numCols);
```
szarnyasg: missing spaces

main/ejml-dsparse/src/org/ejml/sparse/csc/mult/ImplSparseSparseMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,354 @@
...
+            int idxC1 = C.col_idx[colC + 1];
+
+            for (int i = idxC0; i < idxC1; i++) {
+                C.nz_values[i] = x[C.nz_rows[i]];
+            }
+        }
+    }
+
+    /**
+     * Performs the performing operation x = x + A(:,i)*alpha
```
szarnyasg: "performing operation"?

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixVectorMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,124 @@
...
+                Arguments.of("MIN, TIMES", DSemiRings.MIN_TIMES,
+                        new double[]{0.1, Double.MAX_VALUE, 0.2, Double.MAX_VALUE, Double.MAX_VALUE, Double.MAX_VALUE, Double.MAX_VALUE}),
+                Arguments.of("MAX, MIN", DSemiRings.MAX_MIN,
+                        new double[]{0.2, Double.MIN_VALUE, 0.5, Double.MIN_VALUE, Double.MIN_VALUE, Double.MIN_VALUE, Double.MIN_VALUE})
+        );
+    }
+
+    private static Stream<Arguments> matrixVectorMultSources() {
+        return Stream.of(
+                Arguments.of("Plus, Times", DSemiRings.PLUS_TIMES, new double[]{0.5, 0.6, 0, 0, 0, 0, 1.1}),
```
szarnyasg: Why not `"PLUS, TIMES"` here and in other usages of `Arguments`?

main/ejml-core/src/org/ejml/ops/DMonoid.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ejml.ops;
+
+public class DMonoid {
+    // possible re-interpreting 0
```
szarnyasg: This is somewhat difficult to understand. I think what it means is that it is possible to define a different identity element (zero value) instead of `0.0d`. Right?

szarnyasg (changes requested):
Looks really good overall. I've added a few nits inlined.

I've noticed that the signatures of the element-wise operators have changed slightly, e.g. from
```java
    public static void add(double alpha, DMatrixSparseCSC A, double beta, DMatrixSparseCSC B, DMatrixSparseCSC C,
                           @Nullable IGrowArray gw, @Nullable DGrowArray gx)
```
to:
```java
    public static DMatrixSparseCSC add(double alpha, DMatrixSparseCSC A, double beta, DMatrixSparseCSC B, @Nullable DMatrixSparseCSC output,
                                       @Nullable IGrowArray gw, @Nullable DGrowArray gx)
```
They seem to be backwards-compatible with the old signatures so this should not cause a problem. However, I'm wondering whether there was any particular reason to rename the fifth argument from `C` to `output`?

FlorentinD:
@szarnyasg Thank you, for taking the time to review my PR!
 The signature change is based on the changes in 7983443922b4887b9c73d4d525554c8758f6f5f8. There it is also renamed to `output`.

main/ejml-core/src/org/ejml/ops/DMonoid.java
FlorentinD: Right. Maybe I could just mention `neutral element` as a synonym? At least in my linear algebra class, we more often used `neutral element` than `identity`.

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixVectorMultWithSemiRing_DSCC.java
FlorentinD: Good point. Will make it more consistent with the other arguments :+1:

main/ejml-dsparse/src/org/ejml/sparse/csc/mult/ImplSparseSparseMultWithSemiRing_DSCC.java
FlorentinD: this java-doc is just copied, but I agree it could be simplified to `Performs the operation`

main/ejml-core/src/org/ejml/ops/DMonoid.java
szarnyasg: Yes, IMHO neutral element is definitely a good term to use: https://en.wikipedia.org/wiki/Identity_element
The GraphBLAS docs use "identity element" exclusively, so maybe it's good to mention both.

main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOpsWithSemiRing_DSCC.java
FlorentinD: this java-doc is just copied. I will fix it in both places

main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOpsWithSemiRing_DSCC.java
FlorentinD: auto-format to the rescue :)

main/ejml-core/src/org/ejml/ops/DMonoid.java
FlorentinD: Done for the files I added

szarnyasg (approved):
Thanks, looks all good now.

main/autocode/src/org/ejml/GenerateJavaCode32.java
lessthanoptimal: Yep it would involve a lot of automatically generated code. I think double -> float is handled fairly elegantly now. Would we also need to generate integer data types? I definitely do miss C++ style templates when working in Java land. Although the main thing I dislike about modern C++ is how everything should be a template, making the code pointlessly complex, but let's not get into that discussion here ;-)

main/ejml-core/test/org/ejml/ops/TestAlgebraicStructures.java
```diff
@@ -0,0 +1,57 @@
...
+
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestAlgebraicStructures {
+
+    @Test
+    public void testPreDefinedMonoids() {
```
lessthanoptimal: Nit: In JUnit 5 you don't need to name test functions with test in the function name so it's redundant and they should just describe what is being tested. The functions don't even need to be public. EJML has a mixture of junit 4 and junit 5 style test classes for historical reasons. It's recommended that new tests be package protected and omit `test` from function names.

main/ejml-dsparse/src/org/ejml/sparse/csc/mult/ImplSparseSparseMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,354 @@
...
+import org.ejml.sparse.csc.CommonOps_DSCC;
+
+import javax.annotation.Nullable;
+
+import static org.ejml.UtilEjml.adjust;
+
+/**
+ * based on ImplSparseSparseGraphMult_DSCC
+ */
+public class ImplSparseSparseMultWithSemiRing_DSCC {
```
lessthanoptimal, line 34: Not for this PR, but how difficult do you think it would be to auto generate this from the matrix multiply code? Here's an example of how i added concurrency in another library:

https://github.com/lessthanoptimal/BoofCV/blob/1552a49de2fcbec3dba68b2816f9c6facd16de42/main/boofcv-ip/src/main/java/boofcv/alg/filter/binary/impl/ImplBinaryImageOps.java#L82

Maybe something similar could be done here?

main/ejml-dsparse/test/org/ejml/sparse/csc/misc/TestImplCommonOpsWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.Arrays;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class TestImplCommonOpsWithSemiRing_DSCC {
+
+    @ParameterizedTest
```
lessthanoptimal, line 36: I've been doing this test pattern manually. I'll doing more `@ParameterizedTest` in the future.

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
lessthanoptimal: From this discussion it sounds like some functions haven't been tested? @FlorentinD could you create a new ticket to exhaustively test this class in the future, if that's the case?

I've gotten around that problem through heavy use of reflections. In the unit test I've code up a generic and inefficient implementation that can handle all permutations, then use reflections and a consistent naming scheme to call all functions and compare against the generic function.

lessthanoptimal: As for dense semi rings. Are those of much practical value or is it just a competition thing? As an example, for a geometry problem I coded up a dense and sparse solve and never use the dense implementation. If the problem is small enough that I could use dense it's so small that the sparse solver is very fast.

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,141 @@
...
+                Arguments.of("MAX, PLUS", DSemiRings.MAX_PLUS, new double[]{0.7, 1.1}),
+                Arguments.of("MIN, TIMES", DSemiRings.MIN_TIMES, new double[]{0.1, 0.2}),
+                Arguments.of("MAX, MIN", DSemiRings.MAX_MIN, new double[]{0.2, 0.5})
+        );
+    }
+
+    private static Stream<Arguments> sparseMatrixSources() {
+        Random rand = new Random(42);
+        DMatrixSparseCSC sparseMatrix = RandomMatrices_DSCC.rectangle(10, 10, 15, rand);
+        DMatrixSparseCSC denseMatrix = RandomMatrices_DSCC.rectangle(10, 10, 90, rand);
```
lessthanoptimal, line 134: By passing in the same matrix in `sparse sparse` and `dense dense` cases could you be testing a very trivial case accidentally? I do think that's a good test case but it also might be worth having another different sparse and dense matrix. Also calling a CSC matrix dense threw me for a second.

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixVectorMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,124 @@
...
+    public void mult_A_v(String desc, DSemiRing semiRing, double[] expected) {
+        // graphblas == following incoming edges of source nodes
+        double[] v = new double[7];
+        Arrays.fill(v, semiRing.add.id);
+        v[3] = 0.5;
+        v[4] = 0.6;
+
+        System.out.println("input vector = " + Arrays.toString(v));
+
+        //input.print();
```
lessthanoptimal: zap

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixVectorMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,124 @@
...
+    public void mult_v_A(String desc, DSemiRing semiRing, double[] expected) {
+        // graphblas == following outgoing edges of source nodes
+        double[] v = new double[7];
+        Arrays.fill(v, semiRing.add.id);
+        v[3] = 0.5;
+        v[5] = 0.6;
+
+        System.out.println("input vector = " + Arrays.toString(v));
+
+        //input.print();
```
lessthanoptimal: zap/delete commented out debugging code

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixVectorMultWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.Arrays;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class TestMatrixVectorMultWithSemiRing_DSCC {
```
lessthanoptimal, line 34: @breandan and @szarnyasg you two looked at his unit tests and think they do a good job exercising the code? At a high level they look good to me but figured you two have more domain knowledge.

main/ejml-dsparse/test/org/ejml/sparse/csc/misc/TestImplCommonOpsWithSemiRing_DSCC.java
```diff
@@ -0,0 +1,98 @@
...
+    public void testelementWiseMult(DSemiRing semiRing, double[] expected) {
+        // == graph intersection
+        DMatrixSparseCSC matrix = new DMatrixSparseCSC(3, 3, 4);
+        matrix.set(1, 1, 4);
+        matrix.set(1, 2, -2);
+
+        DMatrixSparseCSC otherMatrix = matrix.copy();
+        otherMatrix.set(1, 1, 3);
+        otherMatrix.set(1, 2, 1);
+
```
lessthanoptimal, line 67: a complete nit, but these extra spaces annoy me. I do it all the time too, then I spend the next few years staring at the code resisting the urge to fix it unless I just happen to need to edit the function for some other reason. I rarely modify code for style reasons because it makes debugging regressions much more difficult.

lessthanoptimal (commented):
This is my first pass. So far looks good to me! Going to go over it again and maybe pick a couple of functions and attempt to learn something about semi rings. There are two other PR's ahead of this one to get merged in. I'm anticipating some merge conflicts, but they shouldn't be bad. `@Nullable` will soon be strictly enforced in the public API via static checks.

main/ejml-core/src/org/ejml/ops/DMonoid.java
lessthanoptimal: When the literature can't make up it's mind I tend to at least add a JavaDoc comment with all the commonly used terms so that someone searching the code can find what they are looking for.

lessthanoptimal:
@szarnyasg For your comment about `C` being changed to `output`, that was an attempt on my part to make it very clear what is input and output. Something Java (or really most languages) don't do a good job of. It does muddy up the math equations that explain what the function does a bit. Any strong opinions on that?

lessthanoptimal:
> Looks good to me! The only other thing I can think of is maybe going fully branchless (if possible). I noticed there are several conditionals in the multiplication inner loops. Not an expert but I've read this can have a significant performance impact.

That is definitely true about removing branches speeding things up. Optimizing sparse matrix code is actually quite difficult. I made a few "improvements" over CSparse, most of which you can't notice because 99% of the CPU is spent in other parts of the code. I'm all for any speed improvements we just need to make sure they are worth the added code complexity.

I think you also made some comment about `Math.abs()` being faster than the manual version found in the code but can't find that comment. Historically it has been found that switching to functions like that have a very small speed up on some platforms and a large decrease in others. On embedded platforms JVM's tend to be really dumb. Plus you have Android which has a completely different VM that based on my experience isn't as good as Oracles. The benchmarking code in EJML is out of date and should be updated (better coverage and better approaches) then we could quickly test these ideas on a bunch of different platforms. In the experimental package there's code which tests different data structures and functions.

EJML is also stuck using Java 1.8 byte code and API for the foreseeable future, but Java 14 syntax is possible. Android and a bunch of the java world requires 1.8. The one time I accidentally uploaded newer byte code jars to Maven Central I got emails telling me how everything was now broken.

lessthanoptimal:
Merged in the other PR's just now and this one is up next. I'll do another pass at the code this evening. Thanks for being patient with the review process!

FlorentinD:
@lessthanoptimal  I will look at your comments tomorrow.

szarnyasg:
@lessthanoptimal 
> @szarnyasg For your comment about C being changed to output, that was an attempt on my part to make it very clear what is input and output. Something Java (or really most languages) don't do a good job of. It does muddy up the math equations that explain what the function does a bit. Any strong opinions on that?

No strong opinions from me. The GraphBLAS API uses `C` but this is a GrB library so `output` is fine.

lessthanoptimal:
> > @szarnyasg For your comment about C being changed to output, that was an attempt on my part to make it very clear what is input and output. Something Java (or really most languages) don't do a good job of. It does muddy up the math equations that explain what the function does a bit. Any strong opinions on that?
> 
> No strong opinions from me. The GraphBLAS API uses `C` but this is a GrB library so `output` is fine.

Is "GrB" a type-o? Well if anyone has a strong opinion it can be undone easily before the next release. Most of it auto generated code.

szarnyasg:
@lessthanoptimal 
> Is "GrB" a type-o? 
I wanted to say that "this *isn't* a GrB [GraphBLAS] library".

The GraphBLAS API's naming convention is `C = A*B` (plus bit for masking and the accumulator operator).

image

But I'm also perfectly fine with `output`.

main/ejml-core/test/org/ejml/ops/TestAlgebraicStructures.java
FlorentinD: Good to know :)

main/ejml-dsparse/src/org/ejml/sparse/csc/mult/ImplSparseSparseMultWithSemiRing_DSCC.java
FlorentinD: You would need to:
* change the method signature + add javadoc for the semiring parameter
* replace the `0` with `semiring.add.id` for init values
* add the semiring parameter to function calls f.i. `multAddColA`
* rewrite `sum + x[rowA] * A.nz_values[ai]` to `semiRing.add.func.apply(sum, semiRing.mult.func.apply(x[rowA], A.nz_values[ai]))` .. guess the hardest part will be when there are `*` and `+` operators you dont wont to replace as they are used for calculating the correct index. However you could store those in a seperate variable for simple code replacement.

I think in summary, this shouldnt be very difficult. For my use-case where I also add `masks` and `accumulators`, making it GraphBLAS like, I prefered to have my non generated files to work with (already thought of auto-generating the code).

main/ejml-dsparse/test/org/ejml/sparse/csc/misc/TestImplCommonOpsWithSemiRing_DSCC.java
FlorentinD: also saw this in your tests and hoped you would like this annotation :)

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
FlorentinD: I will add another sparse and dense matrix.
I hope calling it `denseCSC`, will minimize the confusion.

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixVectorMultWithSemiRing_DSCC.java
FlorentinD: Sure thing. I will also remove the prints in the tests

main/ejml-dsparse/test/org/ejml/sparse/csc/misc/TestImplCommonOpsWithSemiRing_DSCC.java
FlorentinD: I removed the extra line :)

main/ejml-dsparse/test/org/ejml/sparse/csc/mult/TestMatrixMatrixMultWithSemiRing_DSCC.java
FlorentinD: I think the testing can still be more exhaustive, like the `add` Variants of `mult` arent directly tested yet. I think the existing test cover the most vital code paths though.  
I would like to create a ticket for more extensive testing (probably after the PR is merged?).

FlorentinD: As the semi ring is used on primitve operations I think they can also provide a practical value for dense variations. 
Maybe its enough to wait for someone to open an issue for dense ops or handle it by auto-generating the semi ring version (like you already proposed).

FlorentinD:
@lessthanoptimal I rebased the PR onto the current master. It looks like there is a problem with the generation of `EjmlVersion`.

lessthanoptimal:
> @lessthanoptimal I rebased the PR onto the current master. It looks like there is a problem with the generation of `EjmlVersion`.

SNAPSHOT is passing CI and works locally. Could you be more specific? The changes to EjmlVersions are that another field was added and that annoying error message should be fixed.

main/ejml-dsparse/src/org/ejml/sparse/csc/mult/ImplSparseSparseMultWithSemiRing_DSCC.java
lessthanoptimal: yep I tend to start with regular code until maintaining it becomes more of a burden than writing auto generated code.

lessthanoptimal:
For `EjmlVersion` problem, I just reran your branch in CI and it got past that point. It died before with this error:
```
/home/travis/build/lessthanoptimal/ejml/main/ejml-core/src/org/ejml/EjmlVersion.java:11: error: illegal start of expression public static final int GIT_REVISION = ;
```
I'll look into it. It might need a `flush()` added.

lessthanoptimal:
Rebase off of SNAPSHOT when you have a chance. I *think* that issue is now fixed. Going to re-run CI a few times to make sure it passes.

FlorentinD:
Sorry for not being more specific. Yes I also could run it locally, that's why I thought the issue is CI based. 
Fingers crossed, CI succeeds now.

lessthanoptimal:
@FlorentinD is there much more to do? I'm leaning towards just creating tickets for any known gaps in testing / features.

FlorentinD:
No only more extensive testing for semi-rings. For which I would happily create a ticket.

Other features I would include in a follow up PR :)

FlorentinD:
@lessthanoptimal Is there anything you need me to do for merging this PR?

maxindelicato:
Just chiming in to say that this is very welcome functionality and seems like a good first pass at semirings on sparse matrices. Would love to see this merged along with masks!

lessthanoptimal:
@FlorentinD squash and I'll merge. Just make a ticket for anything that's not quite done yet. Thanks!

FlorentinD:
@lessthanoptimal I squashed some of my commits and created tickets for remaining tasks.
If you want I can also just squash them into a single commit

lessthanoptimal:
Merged! Thanks for the contribution @FlorentinD  and to @breandan and @szarnyasg for reviewing!

PR #76, opened 2020-07-28
Fix reshapeOrDeclare bug for sparse matrices

FlorentinD:
.. previously used version resulted in row and col values being ignored if the output matrix was not declared yet.
This resulted in wrongly shaped result matrices.

The test could be removed, as it is only to illustrate the bug.

### DIFF
--- a/main/ejml-core/src/org/ejml/UtilEjml.java
+++ b/main/ejml-core/src/org/ejml/UtilEjml.java
@@ -112,6 +112,36 @@ public class UtilEjml {
 
     /**
      * If the input matrix is null a new matrix is created and returned. If it exists it will be reshaped and returned.
+     * @param a (Input/Output) matrix which is to be checked. Can be null.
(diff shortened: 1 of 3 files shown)


lessthanoptimal:
There's an even simpler fix:

replace 
`return reference.createLike();`
with
`return reference.create(rows,cols);`

There should be unit tests for all those functions which accept null as output. Probably best done with reflections. I can handle that later.

FlorentinD:
Just thought the existing behavior was used by operations over dense matrices. 
Changing it inside the method might break those.

lessthanoptimal:
cut and paste error. also a good examples for why you should write new unit tests when you add new features.

lessthanoptimal:
just did the 1 line fix. I also expanded unit tests for matrix input shape to sparse code. fixed a few issues I found with the original dense check too. Thanks for fixing it/reporting!

PR #77, opened 2020-08-11
Remove findbugs dependency

genos:
Hi!

First of all, thanks _very_ much for `ejml`. It's a great library with lots of helpful documentation, and I've really enjoyed using it.

At work, we've recently been going through auditing the licenses of our opensource dependencies (and transitive dependencies, it's been quite an exercise). In doing so, we saw that `ejml` relies on `findbugs`. While `findbugs` has been helpful in the past, its current state is less than ideal:

- The maintainer walked away from the GitHub repository without giving anyone else access for over a year, causing an exodus, abandonment, and replacements popping up.
- More pertinent to my work's interests, There’s an [issue on the GitHub page](https://github.com/findbugsproject/findbugs/issues/128) where the licensing information from the repo (BSD) disagrees with what’s available via Maven (Apache). In any event, the GitHub page for the 3.0.2_preview2 branch says it's [BSD licensed](https://github.com/findbugsproject/findbugs/blob/3.0.2_preview2/findbugs/licenses/LICENSE-jsr305.txt). If instead, you go to the Maven repository and [download the JAR](https://repo1.maven.org/maven2/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2.jar), then unpack it, you find the Apache 2.0 license:
```
~/Downloads/jsr305 ∃ find . -type f -exec grep -i license {} +
./META-INF/MANIFEST.MF:Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
./META-INF/maven/com.google.code.findbugs/jsr305/pom.xml:  <licenses>
./META-INF/maven/com.google.code.findbugs/jsr305/pom.xml:    <license>
./META-INF/maven/com.google.code.findbugs/jsr305/pom.xml:      <name>The Apache Software License, Version 2.0</name>
./META-INF/maven/com.google.code.findbugs/jsr305/pom.xml:      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
./META-INF/maven/com.google.code.findbugs/jsr305/pom.xml:    </license>
./META-INF/maven/com.google.code.findbugs/jsr305/pom.xml:  </licenses>
```

We noticed that the only use of `findbugs` in `ejml` is the `@Nullable` annotation. As such, this PR

- removes the `findbugs` dependency from `build.gradle`, and
- removes all instances of `import javax.annotation.Nullable;` and `@Nullable` from the source.

Everything still compiles, tests, and checks via `gradle` just fine:

```
~/github/ejml ∃ ./gradlew autogenerate
~/github/ejml ∃ ./gradlew build
~/github/ejml ∃ ./gradlew test
~/github/ejml ∃ ./gradlew check
```

Hope this helps! And thanks again for your time and for `ejml`.


genos:
I'm also bewildered by the massive line count on this PR; is that due to `./gradlew autogenerate`? All I did was delete the use of `findbugs` and all instances of `Nullable`.

FlorentinD:
@genos yes this is due to `autogenerate`. There you generate the `float` equivalent, based on the `double` implementation.
I think, you shoudn't commit those autogenerated files.

Personally I like the `@Nullable` Annotations. Isn't there a replacement? In another project I saw the use of `org.jetbrains.annotations`.

genos:
Great! I'll close this PR and open another with `org.jetbrains.annotations.Nullable` instead.

As for not including the autogenerated files, should I add things to the `.gitignore` so I don't commit them by accident?

FlorentinD:
That might be useful.
When committing via IntelliJ, I could easily distinguish the files though.

 @lessthanoptimal what do you think?

genos:
It looks like the autogenerated code ends up in a bunch of different directories, so maybe figuring out the correct `.gitignore` rules would be a bit much. In any event, I'll close this PR and open a new one for `org.jetbrains.anntotations.Nullable` shortly.

genos:
Closing in favor of #78. Thanks again!

PR #78, opened 2020-08-12
Replace findbugs annotation with jetbrains

genos:
# Introduction

Per #77, the current state of `findbugs` is less than ideal. As such, this PR replaces it with `java-annotations` [from Jetbrains](https://github.com/JetBrains/java-annotations).

# Changes

First, I replaced `'com.google.code.findbugs:jsr305:3.0.2'` with `'org.jetbrains:annotations:20.0.0'` in `build.gradle`. Then, I changed all imports of `javax.annotation.Nullable` via

```
grep -rl 'import .*Nullable;' . | xargs sed -i '' "s/import javax.annotation.Nullable;/import org.jetbrains.annotations.Nullable;/g"
```

# Testing

Everything still compiles, tests, and checks via gradle just fine:

```
~/github/ejml ∃ ./gradlew autogenerate
~/github/ejml ∃ ./gradlew build
~/github/ejml ∃ ./gradlew test
~/github/ejml ∃ ./gradlew check
```

Thanks very much!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -80,7 +80,7 @@ subprojects {
     }
 
     dependencies {
-        compile 'com.google.code.findbugs:jsr305:3.0.2' // @Nullable
+        compile 'org.jetbrains:annotations:20.0.0' // @Nullable
 
         testImplementation( 'org.junit.jupiter:junit-jupiter-api:5.4.0' )
         testRuntimeOnly( 'org.junit.jupiter:junit-jupiter-engine:5.4.0' )
--- a/main/ejml-core/src/org/ejml/UtilEjml.java
+++ b/main/ejml-core/src/org/ejml/UtilEjml.java
@@ -25,7 +25,7 @@ import org.ejml.interfaces.linsol.LinearSolverSparse;
 import org.ejml.ops.ConvertDMatrixStruct;
 import org.ejml.ops.ConvertFMatrixStruct;
 
-import javax.annotation.Nullable;
+import org.jetbrains.annotations.Nullable;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.text.DecimalFormat;
--- a/main/ejml-core/src/org/ejml/ops/ConvertDArrays.java
+++ b/main/ejml-core/src/org/ejml/ops/ConvertDArrays.java
@@ -22,7 +22,7 @@ import org.ejml.UtilEjml;
 import org.ejml.data.DMatrix4;
 import org.ejml.data.DMatrixRMaj;
 
-import javax.annotation.Nullable;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Converts 1D and 2D arrays to and from EJML data types
--- a/main/ejml-ddense/src/org/ejml/dense/row/CommonOps_DDRM.java
+++ b/main/ejml-ddense/src/org/ejml/dense/row/CommonOps_DDRM.java
@@ -38,7 +38,7 @@ import org.ejml.interfaces.linsol.LinearSolverDense;
 import org.ejml.interfaces.linsol.ReducedRowEchelonForm_F64;
 import org.ejml.ops.DUnaryOperator;
 
-import javax.annotation.Nullable;
+import org.jetbrains.annotations.Nullable;
 import java.util.Arrays;
 
 import static org.ejml.UtilEjml.*;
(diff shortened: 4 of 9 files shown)


lessthanoptimal:
Any reason to prefer IntelliJ's implementation? IntelliJ seems to recognize it just fine and do you know if Eclipse or other IDE users would be adversely affected if changed to IntelliJ's implementation? Been a bit, but I remember when trying to decide which implementation to use this was the closest to a "standard". Apparently the Java standards group couldn't reach a consensus and gave up, so there's been chaos every since.

genos:
First, thanks again for a great matrix library!

Regarding the different implementations of Nullable, my preference is pragmatic rather than technical. As I mentioned in the previous PR I closed (#77), at work we’ve been going over our open source dependencies, and the transitive findbugs dependency raised some red flags for us. The library has been abandoned after some maintainer drama, and there’s confusion over which license exactly to use for the code. I don’t personally have strong feelings, but my employer thinks it makes sense to not depend on findbugs, and they thought a PR to ejml might help adoption in industry for companies with similar licensing concerns.

lessthanoptimal:
You're welcome!

Thanks for pointing out that PR again. Catching up on e-mail after being on vacation without internet for most of last week and didn't see it since it was already closed.

Searching now to see if `org.jetbrains:annotations` is widely accepted or not. As a result of that search I'm going to try to add static analysis to EJML as part of it's build process. Trying to decide which tools to use. Have you used any of those? `SpotBugs`  looks like a strong candidate. Or maybe https://github.com/google/error-prone ?

genos:
I'm a huge fan of static analysis and code quality tools, though my

genos:
(Sorry, hit the wrong button by mistake!)

I'm a fan of static analysis and code quality tools, but I'm less familiar with the Java landscape. It seems that the `jetbrains` one is somewhat widespread, but not as much as the `findbugs` one; see [this extensive SO answer](https://stackoverflow.com/a/42695253) for more. Interestingly, the `jetbrains` one has `@Retention(CLASS)` while the other is `@Retention(RUNTIME)` if that matters to you. Per the same SO link, the [Checker Framework](https://checkerframework.org) could also prove useful, and if you're just using the annotations and utility classes rather than the full framework, it's [MIT licensed](https://checkerframework.org/manual/#license) (which my employer would prefer) compared to the GPL they use for the framework itself.

`SpotBugs` seems to be the successor to `findbugs`, but
1. SpotBugs [migrated off the JSR305 annotations](https://github.com/spotbugs/spotbugs/pull/180) in favor of their own ones, which I think does away with any hope of having a "standardized" one;
2. "Findbugs/spotbugs/jsr305 annotations with package javax.validation.* clash with other modules in Java9+, also possibly violate Oracle license," from [this SO answer](https://stackoverflow.com/a/41763613); and
3. there's a difference between `@Nullable` vs. `@CheckForNull` that seems to elude many people; see [this bug report](https://sourceforge.net/p/findbugs/bugs/1181/) for instance

Google [Error Prone](https://errorprone.info) looks _very_ useful and fully-featured, even beyond this annotation discussion, and incorporating it into the build system doesn't seem to be too much work.

lessthanoptimal:
Spent sometime today trying to turn `@Nullable` into compile time checks instead of hints for those who read the documentation. I was unable to get `SpotBugs` to work with EJML using Gradle after a bit of effort. However, `Error Prone` worked with minimal effort. Caught a couple of minor issues too.

https://github.com/lessthanoptimal/ejml/tree/errorprone

Anyways, I should have some for of compile time null safety up and running soon. After I do that I'll try merging in this PR and I think it should still work based on the messages I've read.

lessthanoptimal:
Got everything working with Error Prone and NullAway. Let me know when you have reviewed all 227 modified files https://github.com/lessthanoptimal/ejml/pull/79  (I'm joking). Next up will be the switch to jetbrain's annotations. That should go smoothly.

Also thanks for the links above. I'm probably incorrectly using `@Nullable` and `@CheckForNull` in the couple functions which do return a null value. I'm not sure if that even happens in the public API. NullAway is mostly used for the public API. It's not really designed to handle private class fields which have a coupled null state or detect if a function is never called if the value could possible be null.

I'll post another message here after I take one last look at the `Error Prone` PR and merge it. Then it will be time for a heavily modified version of this PR. Would you like to have the honor of doing that? I can also just run the one line script you posted.

genos:
Nicely done with the big push! I'd be happy to lend some time to reviewing #79 if there's a particular part you'd like me to look at (all 227 files does seem a bit much though 😄).

I don't need the honor, thanks! Hopefully, the bash one-liner I posted should be clear enough for others to follow. I'm on a Mac, and the `sed` utility's `-i` option [works differently here than on Linux](https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux); per the answer to that post, one could use something more portable like `perl -i` instead.

lessthanoptimal:
Ended up just using Intelij's search and replace. Everything looks good and I verified that the Nullable checks do still work. Thanks!

Issue #82, opened 2020-08-27
Example using Semirings


lessthanoptimal:
Simple example showing how to use Semirings. Preferably not something which is too abstract that you need to be an expert in the field to understand what's going on.

lessthanoptimal:
Any ideas?  @FlorentinD @breandan @szarnyasg @maxindelicato

maxindelicato:
My primary use-case is their use in graph algorithms, of which the most well-known standard is GraphBLAS: [http://graphblas.org/index.php?title=Graph_BLAS_Forum](http://graphblas.org/index.php?title=Graph_BLAS_Forum); also, see here for further actionable details: [https://www.mit.edu/~kepner/GraphBLAS/GraphBLAS-Math-release.pdf](https://www.mit.edu/~kepner/GraphBLAS/GraphBLAS-Math-release.pdf). Perhaps some simple BFS examples could work? Or something along those lines?

FlorentinD:
I agree bfs is a good example. 
I have some first bfs implementations on my fork in the mask branch (working but definitely not perfect) under BFS_DSCC.
As the branch name indicates, these implementations also include masks. 
However they might be a good starting point.

Shortest path algorithm might also be a good choice to show different semirings

FlorentinD:
And a really simple example is basically in my matrix vector test, which are one bfs iteration. E.g the elements in the vector represent the starting nodes and the matrix the adjacency matrix of a graph

PR #83, opened 2020-09-05
Add an example for the usage of different semirings

FlorentinD:
.. as requested in #82

this is just a starting point.
I tried to create a very minimalistic example. 
Maybe someone has some suggestions for better/more explanations in the example.

### DIFF
--- a/examples/src/org/ejml/example/ExampleGraphPaths.java
+++ b/examples/src/org/ejml/example/ExampleGraphPaths.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2009-2020, Peter Abeles. All Rights Reserved.
+ *
+ * This file is part of Efficient Java Matrix Library (EJML).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ejml.example;
+
+import org.ejml.data.DMatrixSparseCSC;
+import org.ejml.ops.DMonoids;
+import org.ejml.ops.DSemiRing;
+import org.ejml.ops.DSemiRings;
+import org.ejml.sparse.csc.CommonOpsWithSemiRing_DSCC;
+
+/**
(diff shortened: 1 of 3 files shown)


examples/src/org/ejml/example/ExampleSemiRings.java
```diff
@@ -0,0 +1,81 @@
...
+package org.ejml.example;
+
+import org.ejml.data.DMatrixSparseCSC;
+import org.ejml.ops.DMonoids;
+import org.ejml.ops.DSemiRing;
+import org.ejml.ops.DSemiRings;
+import org.ejml.sparse.csc.CommonOpsWithSemiRing_DSCC;
+
+/**
+ * Example showing the usage of semirings in the context of graphs as for instance used in GraphBLAS
```
lessthanoptimal: Could you explain very briefly which problem you are solving in the description. Also might want to change the class name if possible. E.g.
`Example showing how to find the lowest cost path in a graph using semirings`

Then `ExampleGraphBestPath`

More people will understand that language and might realize this is useful for them.

examples/src/org/ejml/example/ExampleSemiRings.java
```diff
@@ -0,0 +1,81 @@
...
+
+        System.out.println("Node 3 can be reached from node 0: " + (outputVector.get(0, 3) == 1));
+        System.out.println("Node 1 can be reached from node 0: " + (outputVector.get(0, 1) == 1));
+
+
+        // Add node 3 to the start nodes
+        startNodes.set(0, 3, 1);
+
+        // Find the path with the minimal cost
+        CommonOpsWithSemiRing_DSCC.mult(startNodes, adjacencyMatrix, outputVector, min_times, null, null);
```
lessthanoptimal: How is the minimum cost path defined in this situation. Sum of cost across the entire path?

lessthanoptimal (commented):
A couple of comments about documentation.

examples/src/org/ejml/example/ExampleSemiRings.java
FlorentinD: In this case its just the cost of the relationship between the nodes, as we are just doing one iteration of bfs.
As we are using the `MIN_TIMES` semi-ring its weight in startNodes vector * relationship-cost (but the startNodes entry are both `1` for simplicity).

I hope my added comments are understandable :)

FlorentinD:
@lessthanoptimal I addressed your comments and hope its now more understand-able

lessthanoptimal:
thanks! Approving. Hopefully you can pick this up again in the future and add more functionality!

lessthanoptimal (approved):
Approved but could you squash first before we merge?

FlorentinD:
Squashed.
I plan to add more functionality in general like masks for sparse matrices (which is on my fork for now). However I want to focus on my thesis for now and create PRs in January.

PR #90, opened 2020-09-27
Fix incorrect indexing in readCDRM and readZDRM

dedztbh:
readCDRM and readZDRM in ReadMatrixCsv is using the wrong column index with set. This causes numbers to be set to wrong locations and eventually causes ArrayIndexOutOfBoundsException.


lessthanoptimal:
Thanks for finding that. Was a bit surprised that slipped through then I noticed there are no unit tests for that function. I'll add those.

lessthanoptimal (approved):

PR #91, opened 2020-09-29
Fixed integer over flow bug 

lessthanoptimal:
- Fixed integer over flow bug when defining sparse matrices and creating a random sparse matrix
- Code formatting

### DIFF
--- a/main/ejml-core/src/org/ejml/UtilEjml.java
+++ b/main/ejml-core/src/org/ejml/UtilEjml.java
@@ -608,6 +608,16 @@ public class UtilEjml {
         return workspace;
     }
 
+    /**
+     * Checks to see if a matrix of this size will exceed the maximum possible value an integer can store, which is
+     * the max possible array size in Java.
+     */
+    public static boolean exceedsMaxMatrixSize( int numRows, int numCols ) {
+        if (numRows == 0 || numCols == 0)
+            return false;
+        return numCols > Integer.MAX_VALUE/numRows;
+    }
+
     public static void printTime( String message, Process timer ) {
         printTime("Processing... ", message, timer);
     }
--- a/main/ejml-core/src/org/ejml/data/DMatrixSparseCSC.java
+++ b/main/ejml-core/src/org/ejml/data/DMatrixSparseCSC.java
@@ -36,9 +36,6 @@ import java.util.Iterator;
  * Row indexes must be specified in chronological order.
  * </p>
  *
- *
- * TODO fully describe
- *
  * @author Peter Abeles
  */
 public class DMatrixSparseCSC implements DMatrixSparse {
@@ -93,6 +90,8 @@ public class DMatrixSparseCSC implements DMatrixSparse {
      * @param arrayLength Initial maximum number of non-zero elements that can be in the matrix
      */
     public DMatrixSparseCSC( int numRows, int numCols, int arrayLength ) {
+        if (numRows < 0 || numCols < 0 || arrayLength < 0)
+            throw new IllegalArgumentException("Rows, columns, and arrayLength must be not be negative");
         this.numRows = numRows;
         this.numCols = numCols;
         this.nz_length = 0;
@@ -463,8 +462,8 @@ public class DMatrixSparseCSC implements DMatrixSparse {
 
     @Override
     public Iterator<CoordinateRealValue> createCoordinateIterator() {
-        return new Iterator<CoordinateRealValue>() {
-            CoordinateRealValue coordinate = new CoordinateRealValue();
+        return new Iterator<>() {
+            final CoordinateRealValue coordinate = new CoordinateRealValue();
             int nz_index = 0; // the index of the non-zero value and row
             int column = 0; // which column it's in
 
--- a/main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
+++ b/main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
@@ -47,7 +47,10 @@ public class RandomMatrices_DSCC {
     public static DMatrixSparseCSC rectangle( int numRows, int numCols, int nz_total,
                                               double min, double max, Random rand ) {
 
-        nz_total = Math.min(numCols*numRows, nz_total);
+        if (UtilEjml.exceedsMaxMatrixSize(numRows, numCols))
+            throw new IllegalArgumentException("Due to how a random matrix is created, rows*cols < Integer.MAX_VALUE");
+
+        nz_total = Math.min(numRows*numCols, nz_total);
         int[] selected = UtilEjml.shuffled(numRows*numCols, nz_total, rand);
         Arrays.sort(selected, 0, nz_total);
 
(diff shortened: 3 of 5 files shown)


main/ejml-core/test/org/ejml/data/TestDMatrixSparseCSC.java
```diff
@@ -45,25 +45,30 @@ public boolean isStructureValid(DMatrixSparse m) {
         return CommonOps_DSCC.checkStructure((DMatrixSparseCSC)m);
     }
 
+    @Test
+    public void constructor_tooLarge() {
+        assertThrows(IllegalArgumentException.class,()->new DMatrixSparseCSC(1_000_000_000,1_000_000_000,4));
+    }
+
```
FlorentinD: `10^9` is actually smaller than `Integer.MAX_VALUE`. So this would be actually a valid sparse matrices.

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
```diff
@@ -45,18 +44,21 @@
      * @param rand Random number generator
      * @return Randomly generated matrix
      */
-    public static DMatrixSparseCSC rectangle(int numRows , int numCols , int nz_total ,
-                                             double min , double max , Random rand ) {
+    public static DMatrixSparseCSC rectangle( int numRows, int numCols, int nz_total,
+                                              double min, double max, Random rand ) {
+
+        if (UtilEjml.exceedsMaxMatrixSize(numRows, numCols))
+            throw new IllegalArgumentException("You have specified a matrix which exceeds the size of an integer");
```
FlorentinD: this more like for temporary fix for the random sparse matrix right?
Maybe we also want to assert, that `nz_total` has a positive value?

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
```diff
@@ -90,42 +91,42 @@ public static DMatrixSparseCSC rectangle(int numRows , int numCols , int nz_tota
...
      * @return Randomly generated matrix
      */
-    public static DMatrixSparseCSC symmetric( int N , int nz_total ,
-                                              double min , double max , Random rand) {
+    public static DMatrixSparseCSC symmetric( int N, int nz_total,
+                                              double min, double max, Random rand ) {
 
         // compute the number of elements in the triangle, including diagonal
-        int Ntriagle = (N*N+N)/2;
+        int Ntriagle = (N*N + N)/2;
```
FlorentinD, line 102: we also need to check here for an overflow

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
```diff
@@ -235,13 +237,13 @@ public static int nonzero( int numRows , int numCols , double minFill , double m
      * @param rand random number generator
      * @return Random matrix
      */
-    public static DMatrixSparseCSC triangle( boolean upper , int N , double minFill , double maxFill , Random rand ) {
-        int nz = (int)(((N-1)*(N-1)/2)*(rand.nextDouble()*(maxFill-minFill)+minFill))+N;
+    public static DMatrixSparseCSC triangle( boolean upper, int N, double minFill, double maxFill, Random rand ) {
+        int nz = (int)(((N - 1)*(N - 1)/2)*(rand.nextDouble()*(maxFill - minFill) + minFill)) + N;
```
FlorentinD: here again `(N-1)*(N-1)` could overflow

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
FlorentinD: So maybe we could change the error message to something like `This method cannot create random sparse matrices, where the dense version would not fit`

main/ejml-core/test/org/ejml/data/TestDMatrixSparseCSC.java
FlorentinD: (see other comments)

main/ejml-core/test/org/ejml/data/TestDMatrixSparseCSC.java
```diff
@@ -45,25 +45,30 @@ public boolean isStructureValid(DMatrixSparse m) {
...
     @Test
     public void constructor_veryLarge() {
-        DMatrixSparseCSC a = new DMatrixSparseCSC(1_000_000_000,100_000_000,4);
+        DMatrixSparseCSC a = new DMatrixSparseCSC(1_000_000,1_000,4);
```
FlorentinD: I think this behaviour should still be allowed, as its not wrong to my point of view

main/ejml-core/src/org/ejml/data/DMatrixSparseCSC.java
```diff
@@ -92,15 +89,17 @@ public DMatrixSparseCSC(int numRows , int numCols ) {
      * @param numCols Number of columns
      * @param arrayLength Initial maximum number of non-zero elements that can be in the matrix
      */
-    public DMatrixSparseCSC(int numRows , int numCols , int arrayLength) {
+    public DMatrixSparseCSC( int numRows, int numCols, int arrayLength ) {
+        if (UtilEjml.exceedsMaxMatrixSize(numRows, numCols))
+            throw new IllegalArgumentException("You have specified a matrix which exceeds the size of an integer");
```
FlorentinD: For sparse matrices this is actually valid to do so. The important point is, that the nz-values do not exceed Integer.MAX_VALUE

main/ejml-core/src/org/ejml/UtilEjml.java
```diff
@@ -574,35 +577,44 @@ public static String fancyString(double value, DecimalFormat format, boolean has
...
     }
 
+    /**
+     * Checks to see if a matrix of this size will exceed the maximum possible value an integer can store, which is
+     * the max possible array size in Java.
+     */
+    public static boolean exceedsMaxMatrixSize( int numRows, int numCols ) {
+        if (numRows == 0 || numCols == 0)
+            return false;
+        return numCols > Integer.MAX_VALUE/numRows;
```
FlorentinD, line 618: As mentioned in other comments, I think this check is valid for the current random matrix creation.
However in general a sparse random matrix can have at maximum `numRows == Integer.MAX_VALUE` and `numCols == Integer.MAX_VALUE`. The important thing is, that the nz-values do not exceed this limit.

FlorentinD (changes requested):
TLDR: I think the check should not be applied in general for `DMatrixSparseCSC` but just for the current random matrix generator.

main/ejml-core/src/org/ejml/UtilEjml.java
lessthanoptimal: Isn't that what this is checking? Mathematically this is identical to `numCols*numRows > Integer.MAX_VALUE` but takes in account the possibility of `numCols*numRows` overflowing.

main/ejml-core/test/org/ejml/data/TestDMatrixSparseCSC.java
lessthanoptimal: `nz_rows` can only take on values from 0 to `Integer.MAX_VALUE` if `numRows*numCols` exceeds that value then not every element can be referenced. The long term fix is to change it from int to long, pun intended.

main/ejml-core/src/org/ejml/UtilEjml.java
lessthanoptimal: Never mind I had a brain fart. Fixing all of this.

FlorentinD (approved):
LGTM :+1: 

As previously commented, in the other random-matrix dscc functions like `symmeteric` the check could be also added. But the original bug is resolved this way.

lessthanoptimal:
Added additional checks and also noticed the unit tests were lacking so I improved a couple of those.

PR #92, opened 2020-09-30
feature/concurrent_dscc

lessthanoptimal:
- First pass at adding a concurrent implementation of mult for DSCC

### DIFF
--- a/main/ejml-core/src/org/ejml/concurrency/IntRangeObjectTask.java
+++ b/main/ejml-core/src/org/ejml/concurrency/IntRangeObjectTask.java
@@ -87,7 +87,7 @@ public class IntRangeObjectTask<T> extends ForkJoinTask<Void> {
 			}
 			// process the last segment in this thread
 			int index0 = step*stepLength + min;
-			consumer.accept(workspace.get(0),index0,max);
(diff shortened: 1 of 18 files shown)


lessthanoptimal:
@szarnyasg  @FlorentinD thought you two might be interested in this. Was able to quickly implement a concurrent sparse matrix multiplication. I'm getting about a 1.9x speed up on 4-core computer.

FlorentinD:
That’s sounds awesome :) 

I think it would be very useful, if the user could specify the used concurrency.
Maybe as a global option. 
In this case the system could also decide on its own, which implementation to use.

lessthanoptimal:
That's the plan, but I might wait until the next release. A lot of code to hook that into and tests to make sure it's done right. I would like to have a new stable release out soon so I need to stop doing massive changes to the code.

In a real-world application this new concurrency code speed it up from  3.3 min to 1 min.

FlorentinD:
Looking forward to it:)

lessthanoptimal:
Right now there's only concurrent add(s,s,s) and mult(s,s,s). Going to look into adding concurrent versions of sparse-dense mult(). There might be hope of making QR partially concurrent without a completely new algorithm.

Issue #84, opened 2020-09-21
Integer overflow for random DSCC matrices with a high dimension.


FlorentinD:
When trying to create a random sparse matrix of size 10^6 x 10^6, a bug occurs based on an integer overflow.

Test to validate:
```
    @Test
    public void hugeDimensionalMatrix() {
        int dim = 1_000_000;
        RandomMatrices_DSCC.rectangle(dim, dim, dim * 4,rand);
    }
```

The problematic lines are (in `RandomMatrices_DSCC`): 
```
nz_total = Math.min(numCols*numRows,nz_total);
int[] selected = UtilEjml.shuffled(numRows*numCols, nz_total, rand);
```

Here `numCols*numRows` produces an Integer overflow.

FlorentinD:
To catch the overflow, we could use `Math.multiplyExact(numRows, numCols)`, but I didn't see, how to handle this case.

lessthanoptimal:
How often do you need matrices larger than the integer max? I've been debating what to do here since the size of Java arrays being limited to 31-bits is getting more annoying and the Java development team doesn't seem to want to move quickly on this problem.

FlorentinD:
For graphs it can be quite often, that the adjacency matrix has such dimensions.
A typical benchmark graph is the ldbc social network graph, with 3 Million vertices as scale-factor 1.
For example a matrix sized `46341` x `46341` already produces an overflow here.

However it occurs less often, that the adjacency matrix actually stores as `2^31-1` entries (e.g. number of edges).
Random matrices are mainly used for testing right?
F.i. for benchmarking my implementations, it would be useful if I could create matrices with higher dimensions.

I also think the method should throw, when `nz_total` is larger than `Int.MAX_VALUE` (or smaller than 0).

FlorentinD:
I could also think of having a different kind of interface for generating sparse random matrices, e.g.`generate(int numRows, int numRows, int avgDegree, someEnum degreeDistribution)`.
With degree I mean the average amount of entries per column (maybe we could find a better word).
Possible distributions would be `uniform` and `powerlaw`. 
The reasoning here is, that a powerlaw distribution is common for many real world graphs.

This idea is based on https://neo4j.com/docs/graph-data-science/current/alpha-algorithms/graph-generation/.

lessthanoptimal:
It now checks to see if the shape exceeds the maximum size of an integer using the function below:
```    
   /**
[...]

PR #95, opened 2020-10-04
Add a random sparse matrix generator based on avg nz entries per column

FlorentinD:
based on issue #84.

This is just a starting point and could be extended by generators using f.i. a power-law or gaussian distribution.
The implemented method always uses a uniform distribtution, e.g. each column has exactly `x` non-zero entries

### DIFF
--- a/main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
+++ b/main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
@@ -26,6 +26,7 @@ import org.ejml.dense.row.CommonOps_DDRM;
 import org.ejml.ops.ConvertDMatrixStruct;
 
 import java.util.Arrays;
+import java.util.BitSet;
 import java.util.Random;
 
 /**
@@ -285,6 +286,62 @@ public class RandomMatrices_DSCC {
         ConvertDMatrixStruct.convert(b, out, UtilEjml.TEST_F64);
 
         return out;
+    }
+
+    /**
+     * Creates a random matrix where each column has exactly `nzEntriesPerColumn` non-zero entries.
+     * Compared to {@link #rectangle} this method can generate larger sparse matrices.
+     *
+     * @param numRows Number of rows
+     * @param numCols Number of columns
+     * @param nzEntriesPerColumn Amount of nz-entries per column
+     * @param min Minimum element value, inclusive
+     * @param max Maximum element value, inclusive
+     * @param sorted Whether the nz-entries should be sorted
+     * @param rand Random number generator
+     * @return Randomly generated matrix
+     */
+    public static DMatrixSparseCSC generateUniform(int numCols, int numRows, int nzEntriesPerColumn,
+                                                   double min, double max, boolean sorted, Random rand) {
+        if (nzEntriesPerColumn > numRows) {
+            throw new IllegalArgumentException("numRows must be greater than nzEntriesPerColumn");
+        }
+
+        int nz_total = Math.toIntExact(nzEntriesPerColumn * numCols);
+
+        DMatrixSparseCSC matrix = new DMatrixSparseCSC(numCols, numRows, nz_total);
+        matrix.indicesSorted = sorted;
+
+
+        int[] nz_hist = new int[numCols];
+        Arrays.fill(nz_hist, nzEntriesPerColumn);
+        matrix.histogramToStructure(nz_hist);
+
+        BitSet existingRows = new BitSet(numRows);
+
+        for (int i = 0; i < nz_total; i++) {
+            if (i % nzEntriesPerColumn == 0) {
+                existingRows.clear();
+            }
+
+            int row = rand.nextInt(numRows);
+            // avoid duplicate entries
+            while(existingRows.get(row)) {
+                row = rand.nextInt(numRows);
+            }
+            matrix.nz_rows[i] = row;
+            existingRows.set(row);
+
+            matrix.nz_values[i] = rand.nextDouble() * (max - min) + min;
+        }
+
+        if (sorted) {
+            for (int i = 1; i <= numCols; i++) {
+                Arrays.sort(matrix.nz_rows, nzEntriesPerColumn * (i - 1), nzEntriesPerColumn * i);
+            }
+        }
+
+        return matrix;
     }
 
     /**
--- a/main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
+++ b/main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
@@ -28,6 +28,7 @@ import org.ejml.ops.ConvertDMatrixStruct;
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * @author Peter Abeles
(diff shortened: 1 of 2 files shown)


main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
```diff
@@ -277,6 +278,62 @@ public static DMatrixSparseCSC symmetricPosDef( int width, double probabilityZer
...
+     * @param numRows Number of rows
+     * @param numCols Number of columns
+     * @param nzEntriesPerColumn Amount of nz-entries per column
+     * @param min Minimum element value, inclusive
+     * @param max Maximum element value, inclusive
+     * @param sorted Whether the nz-entries should be sorted
+     * @param rand Random number generator
+     * @return Randomly generated matrix
+     */
+    public static DMatrixSparseCSC generateUniform(int numCols, int numRows, int nzEntriesPerColumn,
```
lessthanoptimal: even though CSC is column major, could you stick to row column ordering in public facing API?

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
```diff
@@ -277,6 +278,62 @@ public static DMatrixSparseCSC symmetricPosDef( int width, double probabilityZer
...
+        BitSet existingRows = new BitSet(numRows);
+
+        for (int i = 0; i < nz_total; i++) {
+            if (i % nzEntriesPerColumn == 0) {
+                existingRows.clear();
+            }
+
+            int row = rand.nextInt(numRows);
+            // avoid duplicate entries
+            while(existingRows.get(row)) {
```
lessthanoptimal: This approach to ensuring there are no duplicates can be extremely slow. Imaging if there is only one value left out of a possible million. That's why the shuffle approach was used earlier. Maybe the thing to do is if the probability of selecting a valid value < 5%, create a new subset of values which remain and randomly select from that.

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
```diff
@@ -277,6 +278,62 @@ public static DMatrixSparseCSC symmetricPosDef( int width, double probabilityZer
         return out;
     }
 
+    /**
+     * Creates a random matrix where each column has exactly `nzEntriesPerColumn` non-zero entries.
+     * Compared to `rectangle` this method can generate larger sparse matrices.
```
lessthanoptimal: `{@link #rectangle}` to reference the function

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
```diff
@@ -277,6 +278,62 @@ public static DMatrixSparseCSC symmetricPosDef( int width, double probabilityZer
...
+
+
+        int[] nz_hist = new int[numCols];
+        Arrays.fill(nz_hist, nzEntriesPerColumn);
+        matrix.histogramToStructure(nz_hist);
+
+        BitSet existingRows = new BitSet(numRows);
+
+        for (int i = 0; i < nz_total; i++) {
+            if (i % nzEntriesPerColumn == 0) {
```
lessthanoptimal: Will this cause it to have a triangle like pattern along the diagonal? Just curious. I'm also considering reviving the code on visualizing matrices.

main/ejml-dsparse/test/org/ejml/sparse/csc/TestRandomMatrices_DSCC.java
```diff
@@ -62,6 +65,21 @@ void uniform() {
         assertTrue(CommonOps_DSCC.checkSortedFlag(a));
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {"true", "false"})
+    void generateUniform(boolean sorted) {
+        DMatrixSparseCSC a = RandomMatrices_DSCC.generateUniform(numRows, numRows, 4, -1, 1, sorted, rand);
+
```
lessthanoptimal, line 91: Could you add a check on the number of entries? Doesn't need to be exact but you know the upper and low bound. Also run through a series of different matrix shapes, but keep the number of total elements relative small so it still runs fast.

main/ejml-dsparse/test/org/ejml/sparse/csc/TestRandomMatrices_DSCC.java
```diff
@@ -62,6 +65,21 @@ void uniform() {
...
+        assertTrue(CommonOps_DSCC.checkStructure(a));
+    }
+
+    @Test
+    void generateSortedUniform() {
```
lessthanoptimal: isn't this covered by the test above?

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
```diff
@@ -293,4 +350,13 @@ public static void ensureNotSingular( DMatrixSparseCSC A, Random rand ) {
             A.set(s[col], col, rand.nextDouble() + 0.5);
         }
     }
+
+    /**
+     *
+     */
+    public enum EntryDistribution{
```
lessthanoptimal: This isn't being used yet correct? Could you hold off on adding it until it's used in code?

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
FlorentinD: Thanks for pointing that out.

I think a better way is to decide whether to chose random values or remove random values. So if more than half of the row indices are to be assigned, it will to remove values.

Removing could be done via a bitset

lessthanoptimal: yes that's a much better idea.

lessthanoptimal: Will the remove function have a complexity of O(N)?

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
FlorentinD: I could change it to `avgEntriesPerRow`.
The thing is, that I can only say with 100% certainty, that there are `n` entries per column.

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
FlorentinD: No this is just the reset per column, as previous entries do not matter. I could have also wrote it by iterating per column, but this way I easily got the corresponding entry in the nz arrays.

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
FlorentinD: Right. I also didnt plan to commit it. Slipped through somehow.

main/ejml-dsparse/test/org/ejml/sparse/csc/TestRandomMatrices_DSCC.java
FlorentinD: This wasnt supposed to be commited. As the enum.

main/ejml-dsparse/src/org/ejml/sparse/csc/RandomMatrices_DSCC.java
FlorentinD: I dont think so.
The row selection is based on a random number generator.
At the worst case, the remove variant has to drop half of the row indices.
Correspondingly, for the select variant, the worst case is to select half of the row indices.

Now the random number generator basically assures, that it wont be `O(N²)` but more close to `O(N)`.
Its a bit vague, but I think its a good apporximation. Maybe we a benchmark could provide more insight (potentially also compared against the `rectangle` version).

FlorentinD:
@lessthanoptimal I extended the testing and rewrote the generator to improve read-ability and adding a `dropRow` version for "dense" sparse matrices.

lessthanoptimal (approved):
Looks good! Could you squash and format?

lessthanoptimal:
I did the squash and rebase myself. Found a couple of issues and fixed them. There were two unused variables and the unit test was broken but easily fixed. `round()` returns a `long` not and `int` but the function arguments required an `int`. I need to check my CI configuration to see why it didn't complain.

FlorentinD:
Thank you!

FlorentinD:
Regarding the `long` and `int` issue:
I think this worked, as JUnit implicitly casted to the correct type (https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-argument-conversion). 
A similar case is:
```
@ParameterizedTest
@ValueSource(strings = {"true", "false"}
void test(boolean flag) {
 ....
}

PR #96, opened 2020-10-04
Use binary search if indices are sorted for `nz_index` in CSC matrix

FlorentinD:
This is very useful, for larger columns as this function now takes `O(ld n)` instead of `O(N)`, when the entries are sorted.

### DIFF
--- a/main/ejml-core/src/org/ejml/data/DMatrixSparseCSC.java
+++ b/main/ejml-core/src/org/ejml/data/DMatrixSparseCSC.java
@@ -216,12 +216,16 @@ public class DMatrixSparseCSC implements DMatrixSparse {
         int col0 = col_idx[col];
         int col1 = col_idx[col + 1];
 
-        for (int i = col0; i < col1; i++) {
-            if (nz_rows[i] == row) {
-                return i;
+        if (this.indicesSorted) {
+            return Arrays.binarySearch(nz_rows, col0, col1, row);
+        } else {
+            for (int i = col0; i < col1; i++) {
+                if (nz_rows[i] == row) {
+                    return i;
+                }
             }
+            return -1;
         }
-        return -1;
     }
 
     @Override


main/ejml-core/src/org/ejml/data/DMatrixSparseCSC.java
```diff
@@ -212,16 +212,20 @@ public double unsafe_get( int row, int col, double fallBackValue ) {
      * @param col column coordinate
      * @return nz_row index or -1 if the element does not exist
      */
-    public int nz_index( int row, int col ) {
```
lessthanoptimal: For better or worse the project code style is  have spaces at the beginning and end of arguments. At some point I'm going to add a CI check. If you use IntelliJ it should be properly with the correct formatting if you tell it to clean up the code.

lessthanoptimal (changes requested):
Other than that minor formatting issue this looks good. Did you profile this at all? Curious how much the speed up would be, besides the theoretical number above.

main/ejml-core/src/org/ejml/data/DMatrixSparseCSC.java
lessthanoptimal: Attempting to enforce a constant style is very recent. I've attempted to configure it to only be mildly opinionated. This file I recently enforced the formatting.

FlorentinD:
I implemented masks for sparse matrix operations, by wrapping a matrix and then calling `isAssigned(row, col)`, to determine if the entry in the result matrix should be computed.
Hence I called `nz_index` very extensively when doing f.i. matrix-matrix multiplication.

My benchmarking was getting the global triangle count in a graph. Which mainly corresponds to one masked matrix-matrix multiplication.
As you can easily see, the old version scaled very poorly.
When profiling it, I saw, that 90% of the time I spent in `nz_index`.
With this change I could get an improvement of up to **400%** depending on the scaling.

```
Benchmark,avgDegree,nodeCount,Score,Error,Units
ejmlGlobalSandia,4,10_000,32.459,0.635,ms/op
ejmlGlobalSandia,4,100_000,3464.068,8.206,ms/op
ejmlGlobalSandia,4,1_000_000,304798.323,897.960,ms/op
ejmlGlobalSandiaNew,4,10_000,6.173,0.616,ms/op
ejmlGlobalSandiaNew,4,100_000,66.592,6.925,ms/op
ejmlGlobalSandiaNew,4,1_000_000,772.734,59.254,ms/op
```

FlorentinD:
Regarding the masks, as I already mentioned in a previous PR:
Its a bit more of work than the Semirings and I expect a lot of review comments.
I definitely plan to contribute my work, when I finished my master thesis.

lessthanoptimal:
Thanks for the benchmark results and cleaning the formatting. It's good that you can expect a bunch of comments since that means there are people interested in it! I also might have more code tooling in place by then. Trying to make this library and others that I work on easier to maintain and others to contribute.

If it helps, I can handle the pedantic details on your PRs until you're done with the thesis.

lessthanoptimal (approved):

FlorentinD:
Yes its definitely motivating, to see people interested. 

Thank you for offer, I might come back to it :slightly_smiling_face:

lessthanoptimal:
Also that benchmarking you did on `multTransA` ended up having a fairly large impact down stream in my camera calibration and structure from motion tools. Optimizing this one scene went from 3 minutes to 3 seconds after I swapped out the inefficient matrix multiplication. What's interesting is that I didn't catch it earlier because when I compared my results against other popular libraries they were comparable. Since it was about what I expected I didn't dig in to it any more. I now want to revisit that comparison and see if i messed something up or if it it's absolutely crushing these other libraries.

Issue #98, opened 2020-10-07
Generating large sparse random matrices seems to be very slow


FlorentinD:
Recently I introduced `generateUniform` for larger matrices (#95).
I should have added a benchmark for larger dimension, as it turns out, that its actually very slow.
For a matrix of `10^5 x 10^5` it took 10s on my machine...

One idea would be to parallelize the matrix creation.
Additionally having a benchmark, which compares `rectangle()` with `generateUniform()` might be useful.
As I need it for my personal benchmarks, I plan to resolve this issue myself.

Maybe someone has some additional ideas on how to improve the current implementation.

ennerf:
Skimming through the linked code, I think two easy wins could be:

(1) change `BitSet` to `boolean[]`. It uses more memory, but the performance should be quite a bit better.

(2) replace `AtomicInteger` with a for loop over the array or a custom Int wrapper class (example below). Atomic* adds a lot of unnecessary synchronization overhead.

```java
static class IntHolder {
  int value;
}

IntHolder i = new IntHolder();
int index = i.value++;
```

lessthanoptimal:
As an lower bound have you tried to see how fast it is to just generate random numbers without checking to see if they are legal? The other thing is to look at the other library that inspired this and use the same approach (don't copy the code though) if it's fast.

FlorentinD:
Initial timing:
```
Dim 1000 avgEntries 4 generateUniform = 112 generateRectangle = 43
Dim 1000 avgEntries 8 generateUniform = 128 generateRectangle = 5
Dim 10000 avgEntries 4 generateUniform = 9835 generateRectangle = 195
Dim 10000 avgEntries 8 generateUniform = 9138 generateRectangle = 90
```

After using boolean[] and normal integer instead atomic:
```
Dim 1000 avgEntries 4 generateUniform = 88 generateRectangle = 31
Dim 1000 avgEntries 8 generateUniform = 75 generateRectangle = 5
Dim 10000 avgEntries 4 generateUniform = 8827 generateRectangle = 185
Dim 10000 avgEntries 8 generateUniform = 8746 generateRectangle = 90
```
-> improved a bit

**Seeing the actual bug: `boolean dropRows = (numRows/(float)nzEntriesPerColumn) > 0.5;`
It should be `boolean dropRows = ((float)nzEntriesPerColumn/numRows) > 0.5;`**

```
Dim 1000 avgEntries 4 generateUniform = 5 generateRectangle = 33
[...]

PR #99, opened 2020-10-07
Fix bug in random matrix generator and add a Benchmark

FlorentinD:
Resulting in the wrong row selection strategy being used, and hence being very very slow.

Corresponding issue: #98 

Other changes are:
- when selecting rows, they are directly written
- use boolean[] instead of BitSet
- use for-each instead for Streaming BitSet

Thanks @ennerf for the last two tips!

### DIFF
--- a/main/ejml-dsparse/benchmarks/src/org/ejml/sparse/csc/BenchmarkRandomMatrices_DSCC.java
+++ b/main/ejml-dsparse/benchmarks/src/org/ejml/sparse/csc/BenchmarkRandomMatrices_DSCC.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2009-2020, Peter Abeles. All Rights Reserved.
+ *
+ * This file is part of Efficient Java Matrix Library (EJML).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ejml.sparse.csc;
+
+import org.ejml.data.DGrowArray;
+import org.ejml.data.DMatrixRMaj;
+import org.ejml.data.DMatrixSparseCSC;
+import org.ejml.data.IGrowArray;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * @author Peter Abeles
+ */
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 2)
+@Measurement(iterations = 5)
+@State(Scope.Benchmark)
+@Fork(value = 2)
+public class BenchmarkRandomMatrices_DSCC {
+
+    @Param({"10000","40000"})
+    private int dimension;
+
+    @Param({"10", "300"})
+    private int avgEntriesPerColumn;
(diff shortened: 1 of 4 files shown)


FlorentinD:
When trying to run the benchmark, I had to comment out some others like `WatchDoubleStepQR` as there are some `Nullable` compile errors.

Besides I was a bit surprised, that these aren't JMH benchmarks as they could provide better measures.

lessthanoptimal:
You stumbled into an older part of the code that hasn't been updated to JMH yet. `ejml-experimental` is where ideas were being tested out and is a bit of a repository of "this failed don't try it again" or does X slow things down or not. It was written before JMH existed or was popular.

The `Nullable` error is weird. Do you get it when you run `./gradlew build` ?  According to the NullAway rule in gradle that code should have been ignored since it was in a `benchmarks` directory.

main/ejml-experimental/benchmarks/src/org/ejml/sparse/BenchmarkRandomMatrixGenerator.java
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+
+package org.ejml.sparse;
+
+import org.ejml.sparse.csc.RandomMatrices_DSCC;
+
+import java.util.Random;
+
+public class BenchmarkRandomMatrixGenerator {
```
lessthanoptimal: Could you move this to `main/ejml-dsparse/benchmarks/src/org/ejml/sparse/csc` and JMH-ify it?

lessthanoptimal: The name should be something like `BenchmarkRandomMatrices_DSCC`

lessthanoptimal (changes requested):

main/ejml-experimental/benchmarks/src/org/ejml/sparse/BenchmarkRandomMatrixGenerator.java
FlorentinD: Sure, I just couldn't find this directory before :)

FlorentinD:
As I built something inside the experimental module, I think it just threw than.
Normally running `./gradlew build` runs fine.

FlorentinD:
However I am also getting similar errors in `ejml-dsparse:compileBenchmarskJava`, when executing the main method of the benchmark.

I helped myself by disabling errorprune in the `build.gradle` for getting first jmh results.

First results:
```
Benchmark                                     (avgEntriesPerColumn)  (dimension)  Mode  Cnt     Score     Error  Units
BenchmarkRandomMatrices_DSCC.generateUniform                     10        10000  avgt   10     5.380 ±   0.176  ms/op
BenchmarkRandomMatrices_DSCC.generateUniform                     10        40000  avgt   10    50.859 ±   5.373  ms/op
BenchmarkRandomMatrices_DSCC.generateUniform                    300        10000  avgt   10   218.111 ±  17.141  ms/op
BenchmarkRandomMatrices_DSCC.generateUniform                    300        40000  avgt   10   885.351 ±  43.038  ms/op
BenchmarkRandomMatrices_DSCC.rectangle                           10        10000  avgt   10    87.043 ±   1.487  ms/op
BenchmarkRandomMatrices_DSCC.rectangle                           10        40000  avgt   10  1171.267 ±   8.975  ms/op
BenchmarkRandomMatrices_DSCC.rectangle                          300        10000  avgt   10   633.921 ±  75.517  ms/op
BenchmarkRandomMatrices_DSCC.rectangle                          300        40000  avgt   10  3595.355 ± 107.327  ms/op
```

FlorentinD:
As a side note:
I shortly tried parallelizing the generation using `EjmlConcurrency.loopBlocks()`, but this was ca. 3x slower than the single threaded version. Potentially I need to have separate arrays per thread and merge them at the end. Although the threads write into exclusive parts of the arrays.

lessthanoptimal (approved):
LGTM

lessthanoptimal:
I'm curious why you're getting those errors and not me. Must have a different configuration in IntelliJ.

Threads:
I would need to see the code to provide specific comments. Did you have a different `Random` instance for each thread? I think (double check) it is thread safe by uses synchronize so you really don't want to use it across threads. If there's anything that grows an array that could be killing performance and should also be looked at carefully.

lessthanoptimal:
could you squash this commit too? Thanks.

FlorentinD:
No I didn't use different instances, this might be a cause right.

I will squash the commits tommorow evening :)

FlorentinD:
Squashed them :) 

Regarding the threads:
I will start another try at some point, but the current performance is sufficient for my use-case.

If its running fine on your machine and CI, than it must be my Intelij.

Issue #101, opened 2020-10-12
Support Broadcasting


FlorentinD:
Broadcasting allows f.i. elementWise multiplication for matrices of different size.
A simple example would be the following:
![image](https://user-images.githubusercontent.com/15693197/95754115-dbfda300-0ca2-11eb-9c00-35bdbfcaa23d.png)

Here the vector is virtually expanded to the size of the matrix, hence avoiding the actual duplication of the vector to match the dimensions of the matrix.
The terminology seems to be introduces by NumPy (https://numpy.org/doc/stable/user/basics.broadcasting.html), but also used by TensorFlow.

**I am curious, if there are others which would be interested in this feature.**
My personal use-case was that I wanted to normalize the entries in a matrix, such that the sum per row is 1.
Hence I use `reduceRowWise()` to get the sum per row and than would need broadcasting for elementWise dividing each entry with the sum of the corresponding row.
At the moment I hard coded the last step in a helper method, but I would be cleaner to just use `elementWiseMult()`

lessthanoptimal:
How about `apply` but provide (row,col,value) as input as an alternative to dividing by row?

FlorentinD:
Apply variants would be another idea.
For performance I would suggest 3 ways variants then: (row, value) ; (col, value) and your suggested (row, col, value).
As I would just need to access one array instead of both in the first 2 variants

FlorentinD:
Using `apply` is also simpler to implement, I will probably go with this approach for now :+1:

lessthanoptimal:
If you don't beat me to it I'll add a PR for this feature today or tomorrow. FYI my plan is to do a new stable release very early in November. I would do it sooner but I don't have access to a computer I test with at the moment.

FlorentinD:
Great, just ping me, if you would like me to review the PR :)

Tomorrow, I think I will add a power law distribution for the random sparse matrices.

FlorentinD:
I had some wrong ideas about the random power law graph generation, and will add it at a later point.

However, I will probably beat you with the apply PR in the next hours :smile:

FlorentinD:
implemented in #103

lessthanoptimal:
Should we close this ticket now?

FlorentinD:
I think the extended apply replaces implicit array expansion. Unless there is someone eager in implementing it for ease of usage, I wouldn't see a reason.

PR #103, opened 2020-10-19
Add column/row aware apply

FlorentinD:
- also fixing a bug when specifying `null` as the input for `apply`

Matching issue: #101

### DIFF
--- a/main/autocode/src/org/ejml/GenerateJavaCode32.java
+++ b/main/autocode/src/org/ejml/GenerateJavaCode32.java
@@ -54,6 +54,8 @@ public class GenerateJavaCode32 extends GenerateCode32 {
         prefix32.add("FUnary");
         prefix64.add("DBinary");
         prefix32.add("FBinary");
+        prefix64.add("IDBinary");
+        prefix32.add("IFBinary");
         prefix64.add("DMonoid");
         prefix32.add("FMonoid");
         prefix64.add("DSemiRing");
@@ -101,6 +103,7 @@ public class GenerateJavaCode32 extends GenerateCode32 {
         converter.replacePattern("DScalar", "FScalar");
         converter.replacePattern("DUnary", "FUnary");
         converter.replacePattern("DBinary", "FBinary");
+        converter.replacePattern("IDBinary", "IFBinary");
         converter.replacePattern("DMonoid", "FMonoid");
         converter.replacePattern("DSemiRing", "FSemiRing");
         converter.replacePattern("ConvertD", "ConvertF");
--- a/main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOps_DSCC.java
+++ b/main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOps_DSCC.java
@@ -28,6 +28,7 @@ import org.ejml.interfaces.decomposition.LUSparseDecomposition_F64;
 import org.ejml.interfaces.linsol.LinearSolverSparse;
 import org.ejml.ops.DBinaryOperator;
 import org.ejml.ops.DUnaryOperator;
+import org.ejml.ops.IDBinaryOperator;
 import org.ejml.sparse.FillReducing;
 import org.ejml.sparse.csc.factory.DecompositionFactory_DSCC;
 import org.ejml.sparse.csc.factory.LinearSolverFactory_DSCC;
@@ -1787,7 +1788,8 @@ public class CommonOps_DSCC {
     public static DMatrixSparseCSC apply( DMatrixSparseCSC input, DUnaryOperator func, @Nullable DMatrixSparseCSC output ) {
         if (output == null) {
             output = input.createLike();
-        } else if (input != output) {
+        }
+        if (input != output) {
             output.copyStructure(input);
         }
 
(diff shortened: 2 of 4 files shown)


FlorentinD:
On another note. Could you be so kind and label my PRs in oktober with `hacktoberfest-accepted`? These would be this one, https://github.com/lessthanoptimal/ejml/pull/95 and https://github.com/lessthanoptimal/ejml/pull/96.

lessthanoptimal:
Added the tag. Do you also want me to add `Hacktoberfest` as a topic? I'm assuming you are going after a t-shirt https://hacktoberfest.digitalocean.com/

main/autocode/src/org/ejml/GenerateJavaCode32.java
```diff
@@ -53,7 +53,9 @@ public GenerateJavaCode32() {
         prefix64.add("DUnary");
         prefix32.add("FUnary");
         prefix64.add("DBinary");
+        prefix64.add("IDBinary");
```
lessthanoptimal: Can you keep the ordering of 64 then 32 name? easier to keep track of what is converted into what.

main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOps_DSCC.java
```diff
@@ -1802,6 +1804,58 @@ public static DMatrixSparseCSC apply( DMatrixSparseCSC input, DUnaryOperator fun
...
+     * This applies a given unary function on every nz row,value stored in the matrix
+     *
+     * B = f(A).   A and B can be the same instance.
+     *
+     * @param input (Input) input matrix. Not modified
+     * @param func Binary function accepting (row-index, value)
+     * @param output (Output) Matrix. Modified.
+     * @return The output matrix
+     */
+    public static DMatrixSparseCSC applyRowWise( DMatrixSparseCSC input, IDBinaryOperator func, @Nullable DMatrixSparseCSC output) {
```
lessthanoptimal: Suggestion: Maybe `applyRowIdx` or similar instead? `applyRow` for brevity? When I read row wise I think it's being feed an entire row at a time. Feel free to disagree and keep it as is.

lessthanoptimal: Also don't forget to format the file. I ended up this weekend trying to figure out how to add lint checks for this and other code maintenance issues.

lessthanoptimal (changes requested):
Two minor issues and one suggestion.

FlorentinD:
If you want to add the topic is up to you. I could understand, if you think there might be too much spamming (the reason why the tag was introduced in the first place).
On the other hand, people might actually do something valueable ;)

main/ejml-dsparse/src/org/ejml/sparse/csc/CommonOps_DSCC.java
FlorentinD: I like the `Idx` addition. Btw the whole idea is basically somewhat similar to `setNumToInd()` in Combinatorial BLAS

main/autocode/src/org/ejml/GenerateJavaCode32.java
FlorentinD: Sure :+1:

lessthanoptimal:
I'll add that topic. Was just making sure that contest was the reason for the issue tag.

main/autocode/src/org/ejml/GenerateJavaCode32.java
FlorentinD: Looking at it now, I really don't know, how it got there...

main/autocode/src/org/ejml/GenerateJavaCode32.java
```diff
@@ -54,6 +54,8 @@ public GenerateJavaCode32() {
         prefix32.add("FUnary");
         prefix64.add("DBinary");
         prefix32.add("FBinary");
+        prefix32.add("IFBinary");
```
lessthanoptimal: So close...

lessthanoptimal (approved):

main/autocode/src/org/ejml/GenerateJavaCode32.java
FlorentinD: Oh wow :sweat_smile: 
Sorted by order ... Now it must be correct!
Maybe introducing a Class-Annotation like `@PrimitiveSpecific(floatName="FBinary")` would be an idea as this list will continue to grow.
Based on the annotated classes, you could fill part of the `prefix32`, `prefix64` and `converter.replacePattern`.

lessthanoptimal: That's not a bad idea. The entire project can be scanned fairly quickly for that annotation. Although it might not be an annotation in the end but a comment. The auto code generator isn't part of javac and requiring an added dependency for the annotation, which is never really used, would be annoying.

lessthanoptimal (approved):

main/autocode/src/org/ejml/GenerateJavaCode32.java
FlorentinD: I will create a ticket and pick it up when I find some spare time. sounds interesting to implement.

lessthanoptimal: Let me know when you get serious. I might have the source parser already implemented here for lint checks https://github.com/lessthanoptimal/BoofCV/tree/SNAPSHOT/main/checks/src/main/java/boofcv . I'm planning on unleashing some of those checks on EJML after I clean the code up a bit and maybe move it into its own project with a few other tools. Although it might be better to see if Error Prone is a better location for these types of checks.

FlorentinD: Ok, I will let you know. Created #105 to keep track of it :) 

Some checks should be definitely doable via Error Prone.

PR #106, opened 2020-10-22
Element Search Location

lessthanoptimal:
Adds a new variant of element(Min,Max,AbsMin,AbsMax) for DDRM that returns the element's location. Sparse matrices will be handled by a different PR and will require more thought.

### DIFF
--- a/main/ejml-core/src/org/ejml/data/ElementLocation.java
+++ b/main/ejml-core/src/org/ejml/data/ElementLocation.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2020, Peter Abeles. All Rights Reserved.
+ *
+ * This file is part of Efficient Java Matrix Library (EJML).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ejml.data;
+
+import lombok.Data;
+
+/**
+ * The row and column of an element in a Matrix
+ *
+ * @author Peter Abeles
+ */
+@Data
+public class ElementLocation {
+    /** Row coordinate of an element */
+    public int row;
+
+    /** Column coordinate of an element */
+    public int col;
+
+    public ElementLocation() {}
+
+    public ElementLocation( int row, int col ) {
+        this.row = row;
+        this.col = col;
+    }
+
+    public void set( ElementLocation loc ) {
+        this.row = loc.row;
+        this.col = loc.col;
+    }
+
+    public void set( int row, int col ) {
+        this.row = row;
+        this.col = col;
+    }
+}
--- a/main/ejml-core/test/org/ejml/data/TestElementLocation.java
+++ b/main/ejml-core/test/org/ejml/data/TestElementLocation.java
@@ -16,38 +16,26 @@
  * limitations under the License.
  */
 
-package org.ejml.dense.row.misc;
+package org.ejml.data;
 
-import org.ejml.UtilEjml;
-import org.ejml.data.DMatrixRMaj;
-import org.ejml.dense.row.RandomMatrices_DDRM;
 import org.junit.jupiter.api.Test;
-
-import java.util.Random;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-/**
- * Mostly blank because most tests are handled inside of {@link org.ejml.dense.row.TestCommonOps_DDRM}
- *
- * @author Peter Abeles
- */
-public class TestImplCommonOps_DDRM {
-
-    Random rand = new Random(234324);
+public class TestElementLocation {
+    @Test
+    void set_row_col() {
+        var loc = new ElementLocation();
+        loc.set(1, 2);
+        assertEquals(1, loc.row);
+        assertEquals(2, loc.col);
+    }
 
     @Test
-    public void extract() {
-        DMatrixRMaj A = RandomMatrices_DDRM.rectangle(5, 5, 0, 1, rand);
-
-        DMatrixRMaj B = new DMatrixRMaj(3,3);
-
-        ImplCommonOps_DDRM.extract(A, 1, 2, B, 1, 0,2,3);
-
-        for( int i = 1; i < 3; i++ ) {
-            for( int j = 2; j < 5; j++ ) {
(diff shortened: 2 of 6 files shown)


lessthanoptimal:
Performance AFTER

```Benchmark                              (size)   Mode  Cnt      Score       Error   Units
BenchmarkCommonOps_DDRM.elementMax          5  thrpt    5  89620.156 ± 10593.334  ops/ms
BenchmarkCommonOps_DDRM.elementMax       1000  thrpt    5      1.835 ±     0.020  ops/ms
BenchmarkCommonOps_DDRM.elementMaxAbs       5  thrpt    5  73561.357 ±   848.430  ops/ms
BenchmarkCommonOps_DDRM.elementMaxAbs    1000  thrpt    5      1.740 ±     0.150  ops/ms
BenchmarkCommonOps_DDRM.elementMin          5  thrpt    5  63095.981 ±   615.287  ops/ms
BenchmarkCommonOps_DDRM.elementMin       1000  thrpt    5      1.793 ±     0.321  ops/ms
BenchmarkCommonOps_DDRM.elementMinAbs       5  thrpt    5  71645.863 ±   903.977  ops/ms
BenchmarkCommonOps_DDRM.elementMinAbs    1000  thrpt    5      1.777 ±     0.091  ops/ms
BenchmarkCommonOps_DDRM.elementMax          5   avgt    5     ≈ 10⁻⁵               ms/op
BenchmarkCommonOps_DDRM.elementMax       1000   avgt    5      0.545 ±     0.013   ms/op
BenchmarkCommonOps_DDRM.elementMaxAbs       5   avgt    5     ≈ 10⁻⁵               ms/op
BenchmarkCommonOps_DDRM.elementMaxAbs    1000   avgt    5      0.564 ±     0.058   ms/op
BenchmarkCommonOps_DDRM.elementMin          5   avgt    5     ≈ 10⁻⁵               ms/op
BenchmarkCommonOps_DDRM.elementMin       1000   avgt    5      0.543 ±     0.019   ms/op
BenchmarkCommonOps_DDRM.elementMinAbs       5   avgt    5     ≈ 10⁻⁵               ms/op
BenchmarkCommonOps_DDRM.elementMinAbs    1000   avgt    5      0.571 ±     0.026   ms/op
```

lessthanoptimal:
Performance BEFORE

```Benchmark                              (size)   Mode  Cnt      Score     Error   Units
BenchmarkCommonOps_DDRM.elementMax          5  thrpt    5  80322.482 ± 717.291  ops/ms
BenchmarkCommonOps_DDRM.elementMax       1000  thrpt    5      1.834 ±   0.053  ops/ms
BenchmarkCommonOps_DDRM.elementMaxAbs       5  thrpt    5  81925.935 ± 954.457  ops/ms
BenchmarkCommonOps_DDRM.elementMaxAbs    1000  thrpt    5      1.789 ±   0.020  ops/ms
BenchmarkCommonOps_DDRM.elementMin          5  thrpt    5  64592.490 ± 413.218  ops/ms
BenchmarkCommonOps_DDRM.elementMin       1000  thrpt    5      1.825 ±   0.026  ops/ms
BenchmarkCommonOps_DDRM.elementMinAbs       5  thrpt    5  75327.770 ± 603.468  ops/ms
BenchmarkCommonOps_DDRM.elementMinAbs    1000  thrpt    5      1.782 ±   0.021  ops/ms
BenchmarkCommonOps_DDRM.elementMax          5   avgt    5     ≈ 10⁻⁵             ms/op
BenchmarkCommonOps_DDRM.elementMax       1000   avgt    5      0.543 ±   0.005   ms/op
BenchmarkCommonOps_DDRM.elementMaxAbs       5   avgt    5     ≈ 10⁻⁵             ms/op
BenchmarkCommonOps_DDRM.elementMaxAbs    1000   avgt    5      0.560 ±   0.007   ms/op
BenchmarkCommonOps_DDRM.elementMin          5   avgt    5     ≈ 10⁻⁵             ms/op
BenchmarkCommonOps_DDRM.elementMin       1000   avgt    5      0.544 ±   0.008   ms/op
BenchmarkCommonOps_DDRM.elementMinAbs       5   avgt    5     ≈ 10⁻⁵             ms/op
BenchmarkCommonOps_DDRM.elementMinAbs    1000   avgt    5      0.555 ±   0.003   ms/op```

**Comment by lessthanoptimal** [comment]
Despite doing a bit more work the runtime performance appears to the same to within the tests accuracy, i.e. sometimes better sometimes worse. This justifies using a single function for with and without the location.

**Comment by FlorentinD** [comment]
Just one comment, when seeing these functions:
Did you think of using a general `reduceScalarWithLocation` to minimize duplicate code? I wouldn't expect any l performance decrease.

File: main/ejml-core/src/org/ejml/data/ElementLocation.java [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+    public void setRow( int row ) {
+        this.row = row;
+    }
+
+    public int getCol() {
+        return col;
+    }
+
+    public void setCol( int col ) {
+        this.col = col;
```
FlorentinD: why having getters and setters, if the fields are public?
Personally, for this class I would just remove these getters and setters.

FlorentinD (approved):
LGTM.
You could just think about, if you really need getter/setters in the ElementsLocation (see comment)

lessthanoptimal:
I've been thinking of doing just that. What's holding it back is an easy way to run all benchmarks on a lot of different architectures (JVM and physical) and making sure that older systems behave the same. I should also test on Android devices but that's a whole other mess. In the past I did sweeping changes, only to undo it after finding out it got slower by a significant amount on a different computer.

lessthanoptimal:
Oh there's also the issue of memory allocation. For small dense matrices there's a requirement of no memory allocation unless absolutely unavoidable. Memory allocation kills performance in that situation. I think it might be JVM implementation dependent how lambdas are handled. Is it a new instance each time or does it use the same one all the time?

main/ejml-core/src/org/ejml/data/ElementLocation.java
lessthanoptimal: For serialization and bean support. This came across to me as a data structure someone would try to serialize. I was going to add the Lombok `@Data` annotation and let it handle that mess then I noticed this project doesn't have Lombok. Speaking of Serialization it's still missing boiler plate. Maybe I'll just add Lombok.

FlorentinD:
That's an understandable concern.

I would hope it uses the same every time.
Also read something about it a while ago. 
I will put the link in here, when I can find it again.

main/ejml-core/src/org/ejml/data/ElementLocation.java
FlorentinD: Makes sense, thank you for reasoning.

I don't have a strong opinion about Lombok. If its the only place, than I am in favour for explicit getters/setters.

lessthanoptimal: I think I'm going to add it since there are other places i wanted to make more concise. It's funny, I think until Python's popularity sky rotted a few years ago most people were absolute zealots about encapsulation and would complain that I wasn't using accessors everywhere, even if there was no benefit.

PR #108, opened 2020-10-29
Simplify Autocode

lessthanoptimal:
- Made the prefix naming convention more consistent
- Reduced number of exception required to generate code

Just look at autocode files when reviewing to see what really changed.

### DIFF
--- a/main/autocode/src/org/ejml/GenerateJavaCode32.java
+++ b/main/autocode/src/org/ejml/GenerateJavaCode32.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2020, Peter Abeles. All Rights Reserved.
+ * Copyright (c) 2020, Peter Abeles. All Rights Reserved.
  *
  * This file is part of Efficient Java Matrix Library (EJML).
  *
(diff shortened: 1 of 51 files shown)


main/autocode/src/org/ejml/GenerateKotlinCode32.java
```diff
@@ -85,7 +85,7 @@ public GenerateKotlinCode32() {
         }
 
         converter.replacePattern("DScalar", "FScalar");
-        converter.replacePattern("ConvertD", "ConvertF");
```
FlorentinD, line 88: Seeing there is also a kotlin generator, I think I might have missed adding the `DOperator` prefix + converter there.
Or is this not necessary, e.g. how do these lists differ?

main/autocode/src/org/ejml/GenerateJavaCode32.java
```diff
@@ -50,12 +50,8 @@ public GenerateJavaCode32() {
 
         prefix64.add("DGrow");
         prefix32.add("FGrow");
-        prefix64.add("DUnary");
-        prefix32.add("FUnary");
-        prefix64.add("DBinary");
-        prefix32.add("FBinary");
-        prefix64.add("IDBinary");
-        prefix32.add("IFBinary");
+        prefix64.add("DOperator");
```
FlorentinD, line 53: Good idea :+1: 
IMO maybe a bit harder to read, but worth saving the prefixes.

FlorentinD (approved):
LGTM :)

main/autocode/src/org/ejml/GenerateKotlinCode32.java
lessthanoptimal: Kotlin right now is just extensions. Not needed. Most users will probably never even need to type out `DOperator` since it will be implicit as a lambda.

main/autocode/src/org/ejml/GenerateJavaCode32.java
lessthanoptimal: I almost went the suffix route, e.g. F64. Very good discrimination and removes most of those prefixes and no need for new suffices. Everyone thinks they are ugly. What saved that from happening is that there is no standardized suffix for complex variables.

Issue #110, opened 2020-11-09
Wrong error message in LinearSolverQrHouseCol_FDRM/_DDRM


CLOVIS-AI:
The using `LinearSolverQrHouseCol_FDRM` with matrices with the wrong shape, the following exception is thrown:

    Unexpected dimensions for X: X rows = 4 expected = 4

This didn't make sense to me so I took a look into the code, and I found these two lines:
https://github.com/lessthanoptimal/ejml/blob/5788e34695b5e0430c77552cd7c98701a7d31bc6/main/ejml-ddense/src/org/ejml/dense/row/linsol/qr/LinearSolverQrHouseCol_DDRM.java#L117-L118

Maybe it should be:

```java
if( B.numRows != numRows )
    throw new IllegalArgumentException("Unexpected dimensions for B: B rows = "+B.numRows+" expected = "+numRows);
```
Since the test is on `B` and not on `X`.

The error for me happened in the `_FDRM` version (line 113-114) but I didn't find it in the repository, however the `_DDRM` version (linked above) seems to have the same issue.

lessthanoptimal:
I think this is correct:
```
A*X=B
(m*n)(n*k)=(m*k)
```
If `A` is an `m` by `n` matrix, then `B` should have `m` rows. `X` is what is solved for and is determined entirely by the shape of `A` and `B`. In the code you can see it being reshaped to ensure that it is correct.

Could you provide a simple example of what you are doing?

FlorentinD:
I would argue, that `X` is the output and therefore not the `IllegalArgument` as it can be reshaped as you said. The `IllegalArgument`, the parameter with the wrong dimensions, is therefore `B`.

But I also see your point now. Maybe rewriting the message that the dimensions of `A` does not match with `B`?

lessthanoptimal:
Yeah this is probably a communication error from a poorly written error message.

CLOVIS-AI:
Sorry for not being clear, the error is clearly written in the correct case, and my code was wrong, the problem is that the error message `Unexpected dimensions for X: X rows = 4 expected = 4` doesn't make any sense and doesn't help at all to find what the problem is, because the problem is that `B` doesn't have the same size as `A`, but the error message says that `X` doesn't have the same size as `A`, and as you can see from my error message, they were the same size in my case.

lessthanoptimal:
I'll make a generic function to apply to all the solvers with the suggested improved error message.

lessthanoptimal:
[...]

PR #111, opened 2020-11-17
Solve now produces a correct and easily understood error message

lessthanoptimal:
In response to https://github.com/lessthanoptimal/ejml/issues/110


FlorentinD (approved):
Nice improvement :+1:

PR #119, opened 2021-01-06
Introduce concept of Masks

FlorentinD:
This implementation was evaluated as part of my thesis.
Besides Semirings, masks are an important concept of GraphBLAS.

Masks are useful to only compute relevant entries in the result. Compared to offsets, masks allow for arbitrary patterns.
For each result entry, it is checked if mask the entry is set in the mask (`Mask::isSet`). 
For sparse results, this is also useful to avoid large intermediate results. An example use-case is masked mxm for Triangle Count.

In this PR, as an example, the mask is added to dense-vector sparse-matrix multiplication.

(*An idea would be to remove the offsets in these operations as an offset could be specified via a mask.
Therefore I would add a Mask not backed by a matrix but a predicate like `idx > 5`. This would simplify the method signature but also be definitly slower.)

Further planned PRs are:
* Masks in sparse operations
* Mask backed by function instead of matrix

In the future I also plan to add an entry iterator for the mask. This will allow to explicitly only compute the entries set in the mask instead of checking each possible result entry. To chose which strategy to use, the mask should also optionally return its number of non-zero entries (relevant for sparse matrices).

I assume @szarnyasg will be interested in this PR.
Maybe also @breandan or @maxindelicato?

### DIFF
--- a/main/autocode/src/org/ejml/GenerateJavaCode32.java
+++ b/main/autocode/src/org/ejml/GenerateJavaCode32.java
@@ -60,6 +60,8 @@ public class GenerateJavaCode32 extends GenerateCode32 {
         prefix32.add("FMonoids");
         prefix64.add("DSemiRings");
         prefix32.add("FSemiRings");
+        prefix64.add("DMask");
+        prefix32.add("FMask");
         prefix64.add("DScalar");
         prefix32.add("FScalar");
         prefix64.add("DMatrix");
@@ -100,6 +102,7 @@ public class GenerateJavaCode32 extends GenerateCode32 {
         converter.replacePattern("DOperator", "FOperator");
         converter.replacePattern("DMonoid", "FMonoid");
         converter.replacePattern("DSemiRing", "FSemiRing");
+        converter.replacePattern("DMask", "FMask");
         converter.replacePattern("DConvert", "FConvert");
         converter.replacePattern("DGrowArray", "FGrowArray");
         converter.replacePattern("DMatrix", "FMatrix");
@@ -140,6 +143,8 @@ public class GenerateJavaCode32 extends GenerateCode32 {
                 "main/ejml-core/test/org/ejml/data",
                 "main/ejml-core/src/org/ejml/ops",
                 "main/ejml-core/test/org/ejml/ops",
+                "main/ejml-core/src/org/ejml/masks",
+                "main/ejml-core/test/org/ejml/masks",
                 "main/ejml-experimental/src/org/ejml/dense/row/decomposition/bidiagonal/"
         };
 
--- a/main/ejml-core/src/org/ejml/masks/DMaskPrimitive.java
+++ b/main/ejml-core/src/org/ejml/masks/DMaskPrimitive.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2021, Peter Abeles. All Rights Reserved.
+ *
+ * This file is part of Efficient Java Matrix Library (EJML).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ejml.masks;
+
+/**
+ * Mask implementation backed by a primitive array
+ */
+public class DMaskPrimitive extends Mask {
+    // Values interpreted as a row-major dense matrix
+    private final double[] values;
+    /**
+     * Number of columns of the wrapped matrix
+     */
+    public final int numCols;
+    /**
+     * Value representing that the entry is not set in the mask
+     */
+    public final double zeroElement;
+
+    public DMaskPrimitive( double[] values, int numCols, boolean negated, double zeroElement ) {
+        // for dense structures they cannot be used for structural masks
+        super(negated);
+        this.values = values;
+        this.numCols = numCols;
+        this.zeroElement = zeroElement;
+    }
+
+    @Override
+    public boolean isSet( int row, int col ) {
+        // XOR as negated flips the mask flag
+        return negated ^ (values[row*numCols + col] != zeroElement);
+    }
+
+    @Override
+    public int getNumCols() {
+        return numCols;
+    }
+
+    @Override
+    public int getNumRows() {
+        return values.length/numCols;
+    }
+
+    @Override
+    public void setIndexColumn( int column ) {}
+
+    @Override
+    public boolean isSet( int index ) {
+        return negated ^ (values[index] != zeroElement);
+    }
+
+    /**
+     * Utility class to build {@link DMaskPrimitive}
+     */
+    public static class Builder extends MaskBuilder<DMaskPrimitive> {
+        private double[] values;
+        private int numCols = 1;
+        private double zeroElement = 0;
+
+        public Builder( double[] values ) {
+            this.values = values;
+        }
+
+        /**
+         * @param numCols Number of columns in the values
+         */
+        public Builder withNumCols( int numCols ) {
+            this.numCols = numCols;
+            return this;
+        }
+
+        /**
+         * @param zeroElement Value to represent the zero-element in the mask
+         */
+        public Builder withZeroElement( double zeroElement ) {
+            this.zeroElement = zeroElement;
+            return this;
+        }
+
+        @Override
+        public DMaskPrimitive build() {
+            return new DMaskPrimitive(values, numCols, negated, zeroElement);
+        }
+    }
+}
--- a/main/ejml-core/src/org/ejml/masks/Mask.java
+++ b/main/ejml-core/src/org/ejml/masks/Mask.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) 2021, Peter Abeles. All Rights Reserved.
+ *
+ * This file is part of Efficient Java Matrix Library (EJML).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ejml.masks;
+
+import lombok.Getter;
+import org.ejml.MatrixDimensionException;
+import org.ejml.data.Matrix;
+
+/**
+ * Mask used for specifying which matrix entries should be computed
+ */
+public abstract class Mask {
+    /**
+     * Whether the mask entries should be negated.
+     * This avoids materializing the actual negated matrix.
+     */
+    @Getter
+    public final boolean negated;
+
+    protected Mask( boolean negated ) {
+        this.negated = negated;
+    }
+
+    /**
+     * @return Whether the matrix entry is set in the mask
+     */
+    public abstract boolean isSet( int row, int col );
+
+    /**
+     * @return Whether the vector entry is set in the mask
+     */
+    public abstract boolean isSet( int idx );
+
+    /**
+     * @return The number of columns of the wrapped matrix
+     */
+    protected abstract int getNumCols();
+
+    /**
+     * @return The number of rows of the wrapped matrix
+     */
+    protected abstract int getNumRows();
+
+    /**
+     * Prints the mask to standard out.
+     **/
+    public void print() {
+        var result = new StringBuilder();
+        for (int row = 0; row < getNumRows(); row++) {
+            for (int col = 0; col < getNumCols(); col++) {
+                result.append(isSet(row, col) ? "+ " : "- ");
+            }
+            result.append(System.lineSeparator());
+        }
+
+        System.out.println(result);
+    }
+
+    /**
+     * For faster access on a specific column (on at a time)
+     * ! Only useful for sparse masks
+     *
+     * @param column column to index
+     */
+    public abstract void setIndexColumn( int column );
+
+    /**
+     * Checks whether the dimensions of the mask and matrix match
+     *
+     * @param matrix the mask is applied to
+     */
+    public void compatible( Matrix matrix ) {
+        if (matrix.getNumCols() != getNumCols() || matrix.getNumRows() != getNumRows()) {
+            throw new MatrixDimensionException(String.format(
+                    "Mask of (%d, %d) cannot be applied for matrix (%d, %d)",
+                    getNumRows(), getNumCols(), matrix.getNumCols(), matrix.getNumCols()
+            ));
+        }
+    }
+}
--- a/main/ejml-core/src/org/ejml/masks/MaskBuilder.java
+++ b/main/ejml-core/src/org/ejml/masks/MaskBuilder.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2021, Peter Abeles. All Rights Reserved.
+ *
+ * This file is part of Efficient Java Matrix Library (EJML).
+ *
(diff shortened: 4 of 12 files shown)


lessthanoptimal:
Can you rebase off of SNAPSHOT? That should enable the new checks.

lessthanoptimal:
Would it make sense to have a second PR which shows a use case for masks in examples? Do you need to add more capabilities first?

FlorentinD:
I planned to add an example once this is merged. 
Actually I plan to built upon the existing semiring example.

FlorentinD:
For now I thought the tests would suit as an use-case example

main/ejml-core/src/org/ejml/masks/DMasks.java
```diff
@@ -0,0 +1,42 @@
...
+public class DMasks {
+    public static PrimitiveDMask.Builder builder( double[] values ) {
+        return new PrimitiveDMask.Builder(values);
+    }
+
+    public static PrimitiveDMask.Builder builder( DMatrixD1 matrix ) {
+        return new PrimitiveDMask.Builder(matrix.data).withNumCols(matrix.numCols);
+    }
+
+    public static MaskBuilder builder( DMatrixSparseCSC matrix, boolean structural ) {
```
szarnyasg: It'd be worth adding a comment here that explains structural
`only looking if the entry is assigned in the source(disregarding the actual stored value)`
(from `SparseStructuralDMask`).

main/ejml-core/src/org/ejml/masks/PrimitiveDMask.java
```diff
@@ -0,0 +1,86 @@
...
+ * limitations under the License.
+ */
+package org.ejml.masks;
+
+public class PrimitiveDMask extends Mask {
+
+    private final double[] values;
+    // alternatively declare extra separate masks for primitive and dense DMatrix
+    private final int numCols;
+    //
```
szarnyasg: empty comment

szarnyasg (approved):
Nice work - looks all good to me. Looking forward to the mxm variant!

FlorentinD:
@lessthanoptimal when you can find the time, I would appreciate your review :slightly_smiling_face:

lessthanoptimal:
this is close to the top of my queue now! sorry for the delay.

main/autocode/src/org/ejml/GenerateJavaCode32.java
```diff
@@ -60,6 +60,14 @@ public GenerateJavaCode32() {
         prefix32.add("FMonoids");
         prefix64.add("DSemiRings");
         prefix32.add("FSemiRings");
+        prefix64.add("DMasks");
```
lessthanoptimal: What do you think about making DMask the prefix for all of these classes so that can reduce the number of special cases? E.g. SparseDMask -> DMaskSparse

main/ejml-core/src/org/ejml/masks/Mask.java
```diff
@@ -0,0 +1,76 @@
...
+
+import org.ejml.MatrixDimensionException;
+import org.ejml.data.Matrix;
+
+/**
+ * Mask used for specifying which matrix entries should be computed
+ */
+public abstract class Mask {
+
+    // useful for sparse matrices, as actual negation would be costly and result in dense masks
```
lessthanoptimal: This should be a JavaDoc comment?

main/ejml-core/src/org/ejml/masks/Mask.java
```diff
@@ -0,0 +1,76 @@
...
+    public abstract boolean isSet( int row, int col );
+
+    public abstract boolean isSet( int idx );
+
+    public abstract int getNumCols();
+
+    public abstract int getNumRows();
+
+    public void print() {
+        StringBuilder result = new StringBuilder();
```
lessthanoptimal: Suggestion: Java 11 syntax is allowed, but with restrictions. Since you can see the data type in this line and type info is redundant you can use `var` if you like. `var` isn't allowed if you have to browse the code to figure out the type.

main/ejml-core/src/org/ejml/masks/PrimitiveDMask.java
```diff
@@ -0,0 +1,87 @@
...
+ * limitations under the License.
+ */
+package org.ejml.masks;
+
+/**
+ * Mask implementation backed by a primitive array
+ */
+public class PrimitiveDMask extends Mask {
+
+    private final double[] values;
```
lessthanoptimal: comments explaining these fields?

main/ejml-core/src/org/ejml/masks/MaskBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ejml.masks;
+
+public abstract class MaskBuilder<MASK extends Mask> {
```
lessthanoptimal, line 25: JavaDoc comments for class and functions?

main/ejml-core/src/org/ejml/masks/PrimitiveDMask.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public PrimitiveDMask( double[] values, int numCols, boolean negated, double zeroElement ) {
+        // for dense structures they cannot be used for structural masks
+        super(negated);
+        this.values = values;
+        this.numCols = numCols;
+        this.zeroElement = zeroElement;
+    }
+
+    @Override
```
lessthanoptimal: Question: I've been experimenting with having `@Override` on the same line as the function. While not standard in java it reduces the number of vertical lines and doesn't make it less readable. Also fairly common in other languages. Thoughts?

main/ejml-core/src/org/ejml/masks/PrimitiveDMask.java
```diff
@@ -0,0 +1,87 @@
...
+    public boolean isSet( int row, int col ) {
+        // XOR as negated flips the mask flag
+        return negated ^ (values[row*numCols + col] != zeroElement);
+    }
+
+    public double getZeroElement() { return zeroElement;}
```
lessthanoptimal: FYI you can use `@Getter` from Lombok here and remove the explicit getter.

main/ejml-core/src/org/ejml/masks/SparseDMask.java
```diff
@@ -0,0 +1,118 @@
...
+import org.ejml.data.IGrowArray;
+import org.jetbrains.annotations.Nullable;
+
+import static org.ejml.UtilEjml.adjust;
+
+/**
+ * Mask implementation backed by a matrix in CSC format
+ */
+public class SparseDMask extends Mask {
+    protected final DMatrixSparseCSC matrix;
```
lessthanoptimal: comments explaining the fields. This is a general comment throughout this PR

main/ejml-core/test/org/ejml/masks/TestPrimitiveDMasks.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Test
+    void testBasedOnDenseMatrix() {
+        int dim = 20;
+        DMatrixRMaj matrix = RandomMatrices_DDRM.rectangle(dim, dim, new Random(42));
+
+        PrimitiveDMask.Builder maskBuilder = DMasks.builder(matrix);
+        PrimitiveDMask mask =  maskBuilder.withNegated(false).build();
+        PrimitiveDMask negated_mask = maskBuilder.withNegated(true).build();
+
```
lessthanoptimal: nit: there's an extra space here

main/ejml-core/src/org/ejml/masks/Mask.java
lessthanoptimal: If use of `@Getter` is applicable here, it will copy the javadoc comment into the accessor so you don't need to document it twice.

main/ejml-core/src/org/ejml/masks/Mask.java
```diff
@@ -0,0 +1,76 @@
...
+public abstract class Mask {
+
+    // useful for sparse matrices, as actual negation would be costly and result in dense masks
+    public final boolean negated;
+
+    protected Mask( boolean negated ) {
+        this.negated = negated;
+    }
+
+    public abstract boolean isSet( int row, int col );
```
lessthanoptimal, line 42: You should comment all these function in the abstract class. Don't need to do it again in the implementation.

main/ejml-core/src/org/ejml/masks/Mask.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+package org.ejml.masks;
+
+import org.ejml.MatrixDimensionException;
+import org.ejml.data.Matrix;
+
+/**
+ * Mask used for specifying which matrix entries should be computed
+ */
+public abstract class Mask {
```
lessthanoptimal, line 27: Should this implement `Matrix`?

lessthanoptimal: you can then remove many of the function below and there are generic unit tests for testing these common functions.

main/ejml-core/src/org/ejml/masks/DMasks.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package org.ejml.masks;
+
+import org.ejml.data.DMatrixD1;
+import org.ejml.data.DMatrixSparseCSC;
+
+/**
+ * Helper class to get the corresponding mask builder based on a matrix or primitive array
+ */
+public class DMasks {
```
lessthanoptimal: This name I actually found a bit confusing. At first I thought it was a mask with double type. It seems to be a factory for creating masks from double typed structures. Maybe add `Factory` to the name?

lessthanoptimal (changes requested):
Several comments. Mostly about naming and documentation.

main/ejml-core/src/org/ejml/masks/DMasks.java
FlorentinD: Makes sense :+1:

main/ejml-core/src/org/ejml/masks/PrimitiveDMask.java
FlorentinD: As its an annotation, I would keep it on a separate line. Out of curiosity, which languages handle override on the same line?

main/ejml-core/src/org/ejml/masks/Mask.java
FlorentinD: I would like to avoid implementing `Matrix` here, as it should only be wrapping an existing matrix but not creating a dedicated matrix.
Also I plan to have a mask variant simply backed by a predicate, f.i., only compute the lower triangle of the matrix.

Having the mask simply as Matrix in the operations like vxm is a viable idea. Other GraphBLAS implementations only create the Mask object internally. But as vectors can be represented as simple primitive arrays, I chose to have the Mask object directly as an input.

main/ejml-core/src/org/ejml/masks/Mask.java
FlorentinD: True, I totally forgot about adding a proper Java-Doc.

main/autocode/src/org/ejml/GenerateJavaCode32.java
FlorentinD: Looks much more consistent :)

main/ejml-core/src/org/ejml/masks/PrimitiveDMask.java
lessthanoptimal: C++ and Kotlin come to mind. In those cases it's part of the language. Java (probably for backwards compatibility) made it an optional annotation. When I was looking into code formatting I notice some formatters explicitly allow this style.

main/ejml-core/src/org/ejml/masks/Mask.java
FlorentinD: Right, I also improved the comment now.

FlorentinD: And `@Getter` is applicable here

main/ejml-core/src/org/ejml/masks/Mask.java
FlorentinD: Looks like a good place for `var` :)

FlorentinD:
@lessthanoptimal thank you for the review. I think I could apply your comments or explain why I did it this way.

main/ejml-core/src/org/ejml/masks/PrimitiveDMask.java
FlorentinD: Isn't the override als optional in C++? So it's a good question why override is an annotation in Java

main/ejml-core/test/org/ejml/masks/TestDMasksPrimitive.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.Random;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestDMasksPrimitive {
+
+    @Test
+    void testBasedOnPrimitiveArrays() {
```
lessthanoptimal: (optional) seems redundant to have test in the function names in your unit tests because of the `@Test` annotation

main/ejml-core/test/org/ejml/masks/TestDMasksPrimitive.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Test
+    void testBasedOnDenseMatrix() {
+        int dim = 20;
+        DMatrixRMaj matrix = RandomMatrices_DDRM.rectangle(dim, dim, new Random(42));
+
+        DMaskPrimitive.Builder maskBuilder = DMaskFactory.builder(matrix);
+        DMaskPrimitive mask =  maskBuilder.withNegated(false).build();
+        DMaskPrimitive negated_mask = maskBuilder.withNegated(true).build();
+
```
lessthanoptimal, line 54: nit: extra space

main/ejml-core/src/org/ejml/masks/DMaskPrimitive.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Mask implementation backed by a primitive array
+ */
+public class DMaskPrimitive extends Mask {
+    // Values interpreted as a row-major dense matrix
+    private final double[] values;
+    // Number of columns of the wrapped matrix
+    private final int numCols;
+    // Value representing that the entry is not set in the mask
```
lessthanoptimal: Since you used `@Getter` could you turn this comment into javadoc? That way the generated getter will have documentation

main/ejml-core/src/org/ejml/masks/DMaskPrimitive.java
```diff
@@ -0,0 +1,99 @@
...
+package org.ejml.masks;
+
+import lombok.Getter;
+
+/**
+ * Mask implementation backed by a primitive array
+ */
+public class DMaskPrimitive extends Mask {
+    // Values interpreted as a row-major dense matrix
+    private final double[] values;
```
lessthanoptimal, line 25: EJML tends to provide access the internal data structures without requiring accessors for objects which are performance critical. Is that the case here? Basically should this be private or public? This same comment can be applied throughout the code.

lessthanoptimal (approved):
A few minor comments. Could you address those then squash? There were a few areas I wanted to dig deeper into but I think it's more important to get this out and used. I've delayed it enough already.

lessthanoptimal:
Could you also do a follow up PR with micro benchmarks? This is for speed regression testing and shouldn't take longer than a second or so to run each test. I need to figure out a good way to make these regression tests public too. Right now it just goes to my in box.

main/ejml-core/test/org/ejml/masks/TestDMasksPrimitive.java
FlorentinD: that's something I really want to avoid. Thank you for spotting :+1:

main/ejml-core/src/org/ejml/masks/DMaskPrimitive.java
FlorentinD: I don't think these should be accessed through the mask, as the user should have direct access to the array used for the mask creation.
Thats why the isn't even a getter for the matrix/array.
I will apply it for the mask parameter like `negated` though :+1:

main/ejml-core/src/org/ejml/masks/DMaskPrimitive.java
FlorentinD: I chose to make the field public too.

PR #118, opened 2021-01-06
Implement `select` operation for csc matrices

FlorentinD:
A useful operator to f.i. (in-place) select only the lower triangle or diagonal of a matrix.
My use-case was the Triangle Count graph algorithm, implemented via a select + masked multiplication (to be added) + reduce.

A further idea would be to allow value-based selectors. I didn't need it, so I left it as a `TODO` for now. A value-based selector could be useful to remove zero entries or entries above/below a given threshold.

PR #121, opened 2021-01-14
Apply spotless

FlorentinD:
I didn't want to have these changes in #118.
To note is, that the `build.gradle` files still has the wrong year in the copyright.

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -1,25 +1,25 @@
 /*
- * Copyright (c) 2020, Peter Abeles. All Rights Reserved.
- *
- * This file is part of Efficient Java Matrix Library (EJML).
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *   http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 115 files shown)


lessthanoptimal:
Something is very wrong. It should only apply it to files that have been
modified relative to snapshot. Seems to work that way on my machines and
CI. I'll double check in a few hours.

FlorentinD:
Ah ok, thats a cool feature to only consider the modified files.
Although it shouldnt be necessary once SNAPSHOT is compliant to spotless right?

FlorentinD:
I rebased from SNAPSHOT and ran `./gradlew spotlessApply`

lessthanoptimal:
I bet your local SNAPSHOT isn't update with the remote SNAPSHOT. Couldn't reference the remote SNAPSHOT since in CI there was no concept of remote.

Most of the time you don't want to apply it to every single file and gradually update files as needed. Spotless primarily acts as a code formatter and checks that your code is compliant. On some projects people change their mind constantly what the best format is you would be constantly modifying 1000's of files.

lessthanoptimal:
I take that back. It does reference `origin/SNAPSHOT`. What's your set up and can you reproduce this with a branch from SNAPSHOT? If so can you paste the commands you typed to create the branch?

FlorentinD:
I couldn't reproduce it on a freshly cloned repository. There nothing changed after applying spotless. So it must be something with my fork it seems. 

The PR can be closed of course :)

FlorentinD:
The error was based on the SNAPSHOT branch on my fork being not up to date. As the gradle task checks `origin/SNAPSHOT` it compared against an older version. 
--> Either locally rename your fork remote or update the SNAPSHOT branch on your fork

PR #124, opened 2021-01-17
Fix `getNumElements` for sparse matrix formats

### LINKED ISSUE #122: DMatrixSparseCSC.getNumElements() returns number of non zero elements
Hi, 
While using EJML I stumbled about something that might be intended, but which I find pretty counter intuitive.

In `DMatrixSparceCSC` the `getNumElements()` method returns `nz_length` (number of non zero elements or internal array size or something, not sure there), while other classes return `numRows*numCols`. The java doc that is inherited from a superclass also states that this is `numRow * numCols`. 

Personally, I think there should be a separate method that returns the number of non zero elements and this method should return what is stated in the superclass javadoc. 

Kind regards,

FlorentinD:
Also removing redundant and unused `getNonZeroCount` in favor of `getNonZeroLength`. Maybe `getNonZeroLength` is to implementation specific and `getNonZeroCount` would be the better choice.

One could argue about the naming still, as the stored entries can actually be zero. So, `getStoredNumberOfElements` might be more readable but also more verbose. Also it would unnecessarily break the API so I refrained from any renaming.

Fixes #122

### DIFF
--- a/main/ejml-core/src/org/ejml/data/DMatrix.java
+++ b/main/ejml-core/src/org/ejml/data/DMatrix.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Peter Abeles. All Rights Reserved.
+ * Copyright (c) 2021, Peter Abeles. All Rights Reserved.
  *
  * This file is part of Efficient Java Matrix Library (EJML).
  *
@@ -15,7 +15,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.ejml.data;
 
 /**
@@ -73,5 +72,7 @@ public interface DMatrix extends Matrix {
      *
      * @return Number of elements in this matrix.
      */
-    int getNumElements();
+    default int getNumElements() {
+        return getNumRows() * getNumCols();
+    }
 }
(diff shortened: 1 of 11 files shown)


FlorentinD:
Went a step further and provided a default computation of `getNumElements` in the last commit.

main/ejml-core/src/org/ejml/data/DMatrix.java
```diff
@@ -73,5 +72,7 @@
      *
      * @return Number of elements in this matrix.
      */
-    int getNumElements();
+    default int getNumElements() {
+        return getNumRows() * getNumCols();
```
lessthanoptimal, line 76: Good idea. I had to double check since I was thinking this feature came after java 8, but nope it's a Java 8 feature.

lessthanoptimal (approved):

lessthanoptimal:
You're right about it technically not always being zero, but I think most spare matrix literature still reefers to them as non-zero since if is zero you should remove them. Anyways, approved and thanks for the PR. I'll let you click the `rebase and merge` button, which I normally do, since it's more typical for the person who created the PR to do it.

FlorentinD:
I often read about explicit and implicit zeros to distinguish both. Anyway that's something for a different time.

lessthanoptimal:
BTW ignore my earlier comment about java 8 syntax. In this project we should be good with Java 11 syntax but the byte code output is Java 8. Calling API functions after java 8 would be bad. I'm not sure CI can catch that since it's build with java 11.

Issue #132, opened 2021-02-18
binary serialization to/from .mat


ennerf:
Creating a separate issue to continue the discussion from #131

I think that a `MatrixIO.loadMatlab(String fileName)` would have to be limited a very specific file layout, e.g., maybe a single matrix variable with a specific name. It'd act similar to `MatrixIO.loadBin(String fileName)` but be interoperable with Matlab, SciPy, Julia etc.

For supporting more complex file structures it'd probably be better just to use MFL directly, e.g., what should the parser do if it finds two matrices inside a cell array inside a struct?

Do you think a 

```Java
public class MatrixIO {
    public static <T extends DMatrix> T loadMatlab(String fileName) throws IOException {
        if(!hasDependency()) {
            throw new IllegalStateException("Missing dependency: add maven coordinates");
        }
        return reflectivelyCallMethod(fileName);
    }
}
```

would be better than just writing a sentence documentation for calling `Mat5Ejml.loadMatlab(String fileName)` (doesn't exist yet, but would have to be implemented in any case) ?

lessthanoptimal:
throw an exception and tell people to use MFL directly?  Alternatively, you could pass in a List<Matrix> and put every read into the list. Similar for `saveMatlab`

lessthanoptimal:
`throw new IllegalStateException("Missing dependency: add maven coordinates. http://project.website.com");` if the dependency isn't there. Then if a data structure that's too complex is encountered then throw another exception saying basically that this is too complex and go to the project website for instructions on how to use MFL.

lessthanoptimal:
I really think any variant on what we have proposed would work well as long as the user knows how to fix the situation without googling.

ennerf:
Sounds good. I don't have time this week, but I'll try to get on this soon-ish.

PR #133, opened 2021-03-02
added save/load matlab using external library

ennerf:
See #132 for discussion. This does the simplest thing and stores/loads a single variable using the default settings (compressed).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -110,6 +110,7 @@ subprojects {
     }
 
     dependencies {
+        compileOnly 'us.hebi.matlab.mat:mfl-ejml:0.5.7'
         compileOnly 'org.projectlombok:lombok:1.18.10'
         compileOnly 'org.jetbrains:annotations:20.0.0' // @Nullable
         compileOnly 'javax.annotation:jsr250-api:1.0'  // @Generated
--- a/main/ejml-core/src/org/ejml/ops/MatrixIO.java
+++ b/main/ejml-core/src/org/ejml/ops/MatrixIO.java
@@ -15,7 +15,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.ejml.ops;
 
 import org.ejml.EjmlVersion;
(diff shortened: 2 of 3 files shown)


lessthanoptimal:
Have you tested this yet in a project that's missing the dependency? Does `MatrixIO` blow up because of the import statements?

ennerf:
`compileOnly` doesn't add it to the test runtime, so the unit test verifies the failing case.

I've used similar patterns a few times, e.g., for choosing a Java 8 vs 11 implementation at runtime. The nested class usages get moved into a separate `.class` file during during compilation, so the Classloader won't try to load the library until the nested class gets used. Doing it this way plays nicer with obfuscation tools like ProGuard (and presumably Graal's native image?), and has no associated performance cost compared to a normally compiled version.

lessthanoptimal:
Sorry for taking a bit to review this. When I run locally on my home computer a unit test is failing. `load_save_matlab_missing_dep_msg()`

```
org.opentest4j.AssertionFailedError: Expected java.lang.IllegalStateException to be thrown, but nothing was thrown.
...
at org.ejml.ops.TestMatrixIO.load_save_matlab_missing_dep_msg(TestMatrixIO.java:101)
```

lessthanoptimal:
I ran into no issues when calling it from another project without your lib.

ennerf:
Interesting. Building/testing it with the Gradle wrapper worked for me just fine. Can you give me any more information about how you were running the tests?

By no issues, do you mean that you ran `saveMatlab` without MFL on the path and it didn't throw any errors?

lessthanoptimal:
Appears to be an intellij issue. Gradle is the official criteria so it passes! Could you just squash the commit then feel free to merge.

lessthanoptimal (approved):
LGTM squash then merge. Thanks!

ennerf:
Sounds good 👍. I don't have write access to do a (squash) merge, so I'm assuming that you'd be doing that?

image

Running the tests in IntelliJ worked for me as well. Maybe it needs to clean and/or re-synchronize the project

lessthanoptimal:
yeah I tried reloading gradle immediately when I had that issue. One difference could be that I have IntelliJ run tests directly and not through Gradle since it's much faster.

PR #136, opened 2021-04-08
Fix id element of the max monoid

FlorentinD:
`Double.MIN_VALUE` is only the smallest positive value. The actual smallest double value is `-Double.MAX_VALUE`.


lessthanoptimal (approved):
I have made this same mistake on several occasions when switching between floats and ints. Wonder if we could a PR for the JDK to add a new better named constant...

FlorentinD:
I mean the problem will persist as the existing `MIN_VALUE` will unlikely be changed. And its hard to come up with a better name than `MIN_VALUE`

lessthanoptimal:
Could add a `@Deprecated` but yeah `MIN` should have been `SMALLEST` and probably have to wait until `Double` is deprecated to really fix this. Within EJML it would be possible to add a checks for using that constant and forces you to use a EJML constant instead.

Issue #139, opened 2021-05-11
class org.ejml.data.DMatrixRMaj cannot be cast to class org.ejml.data.DMatrixSparseCSC


edoardorealini:
Hello,

I am trying to build a DMatrixSparseCSC through a SimpleMatrix in which data is loaded as a 2D array of doubles generated from a png image. My goal is to store the png as a sparse matrix in a binary file. 

When I try to generate the sparse matrix by calling the method simpleMatrix.getDSCC() on an instance of SimpleMatrix I get the follwing exception: 
java.lang.ClassCastException: class org.ejml.data.DMatrixRMaj cannot be cast to class org.ejml.data.DMatrixSparseCSC 

Am I doing something wrong?

Many thanks,
ER

FlorentinD:
`getDSCC` does not convert the inner matrix into a CSC matrix but only tries to cast it. It looks more of a convenience method if you know your internal matrix is of type CSC.

I think you should use `org.ejml.ops.DConvertMatrixStruct#convert(org.ejml.data.DMatrixRMaj, org.ejml.data.DMatrixSparseCSC, double)`. So use `simpleMatrix.get()` for the DMatrixRMaj, e.g., `var sparseMatrix = convert(simpleMatrix.get(), null, 1e-4)`

** I never worked with the SimpleMatrix interface but only directly with the concrete matrix types`

lessthanoptimal:
Adding a nice early to use function to convert the internal matrix type wouldn't be a bad idea. Maybe something like:

```
SimpleMatrix converted = simpleMatrix.convertType(DMatrixSparseCSC.class);
SimpleMatrix converted = simpleMatrix.convertDDRM();
```

Other suggestions are welcomed.

FlorentinD:
@lessthanoptimal you already implemented these methods in `org.ejml.ops.ConvertMatrixType#convert`. I opened a PR to fix the getters by implicitly converting to the required type.

@edoardorealini In the meantime you can simply use either  `org.ejml.ops.ConvertMatrixType#convert` or `SimpleMatrix.convertToSparse()`.

lessthanoptimal:
resolved

PR #140, opened 2021-05-13
Fix simple matrix `getType` methods

FlorentinD:
based on issue #139 

Now implicitly converting to the required type if needed.
Also added a matrix equals for combination of float and double matrices

### DIFF
--- a/main/ejml-core/src/org/ejml/ops/ConvertMatrixType.java
+++ b/main/ejml-core/src/org/ejml/ops/ConvertMatrixType.java
@@ -28,6 +28,7 @@ public class ConvertMatrixType {
     /**
      * Converts a matrix of one data type into another data type. If no conversion is known then an exception
      * is thrown.
+     * The result is always a new matrix instance, even if the matrix already has the desired type.
      *
      * @return The converted matrix
      */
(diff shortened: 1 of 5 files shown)


FlorentinD:
Forgot to run `spotlessJavaApply` again. Maybe we could add a PR template with a checklist, which includes `Run spotlessJavaApply`

lessthanoptimal:
Might have some tweaks. On a trip at the moment but I'll finish review in a few days I think. Thanks for the quick PR!

FlorentinD:
:wave:  a short reminder to review this PR (@lessthanoptimal)

main/ejml-core/src/org/ejml/ops/ConvertMatrixType.java
```diff
@@ -36,12 +35,13 @@
     public static Matrix convert( Matrix matrix, MatrixType desired ) {
         Matrix m = null;
 
+        if (matrix.getType().equals(desired)) {
```
lessthanoptimal: I think the original behavior here was correct, but the JavaDoc needs to be updated to make explicit what the contract is. The logic is that in every case a new instance should be created. That way you know you're not modifying the original and you don't need to check for the special case where its the same matrix type..

lessthanoptimal (changes requested):
One comment and if you disagree we should discuss what the correct behavior should be. Either way the contract needs to be made explicit.

main/ejml-core/src/org/ejml/ops/ConvertMatrixType.java
FlorentinD: Agree, always making a copy is easier to handle for the user

lessthanoptimal (approved):

PR #143, opened 2021-07-06
feature/simple_concurrent

lessthanoptimal:
- Operations for simple matrix will now switch to concurrent implementation when available

- [ ] Your code is covered by tests

### DIFF
--- a/main/ejml-core/src/org/ejml/concurrency/EjmlConcurrency.java
+++ b/main/ejml-core/src/org/ejml/concurrency/EjmlConcurrency.java
@@ -15,7 +15,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.ejml.concurrency;
 
 import org.ejml.data.Matrix;
--- a/main/ejml-core/src/org/ejml/concurrency/EjmlConcurrency.java
+++ b/main/ejml-core/src/org/ejml/concurrency/EjmlConcurrency.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Peter Abeles. All Rights Reserved.
+ * Copyright (c) 2021, Peter Abeles. All Rights Reserved.
  *
  * This file is part of Efficient Java Matrix Library (EJML).
  *
@@ -18,6 +18,8 @@
 
 package org.ejml.concurrency;
 
+import org.ejml.data.Matrix;
+import org.ejml.data.MatrixSparse;
 import pabeles.concurrency.ConcurrencyOps;
 
 /**
@@ -28,6 +30,9 @@ import pabeles.concurrency.ConcurrencyOps;
 public class EjmlConcurrency extends ConcurrencyOps {
 	/** Used to toggle auto matic switching to concurrent algorithms */
 	public static boolean USE_CONCURRENT = true;
+
+	/** Minimum number of elements in a matrix before it will switch to concurrent implementation */
+	public static int ELEMENT_THRESHOLD = 50_000;
 
 	/**
 	 * Sets the maximum number of threads available in the thread pool and adjusts USE_CONCURRENT. If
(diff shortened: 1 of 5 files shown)


main/ejml-core/src/org/ejml/concurrency/EjmlConcurrency.java
```diff
@@ -29,6 +30,9 @@
 	/** Used to toggle auto matic switching to concurrent algorithms */
 	public static boolean USE_CONCURRENT = true;
 
+	/** Minimum number of elements in a matrix before it will switch to concurrent implementation */
+	public static int ELEMENT_THRESHOLD = 50_000;
```
FlorentinD, line 34: Curious, if this number is based on benchmarks or just some estimation. Assumed its also partly operation depended

main/ejml-core/src/org/ejml/concurrency/EjmlConcurrency.java
```diff
@@ -45,4 +49,26 @@ public static void setMaxThreads( int maxThreads ) {
...
+
+	/**
+	 * Returns true if the operation on the matrix should use the concurrent implementation. If the matrix
+	 * is too small it should always use a single threaded implementation since the overhead will slow it down.
+	 */
+	public static boolean useConcurrent( MatrixSparse mat ) {
+		if (!USE_CONCURRENT)
+			return false;
+
+		return mat.getNonZeroLength() > ELEMENT_THRESHOLD;
```
FlorentinD, line 61: NIT: could write it as `return USE_CONCURRENT && (mat.getNonZeroLength() > ELEMENT_THRESHOLD)`. Short-circuit evaluation assures the same effect and its more compact.

(the same applies for the other method below)

main/ejml-simple/src/org/ejml/simple/ops/SimpleOperations_DDRM.java
```diff
@@ -68,17 +69,29 @@ public void fill( DMatrixRMaj A, /**/double value ) {
 
     @Override
     public void transpose( DMatrixRMaj input, DMatrixRMaj output ) {
-        CommonOps_DDRM.transpose(input, output);
+        if (EjmlConcurrency.useConcurrent(input)) {
+            CommonOps_MT_DDRM.transpose(input, output);
+        } else {
+            CommonOps_DDRM.transpose(input, output);
+        }
```
FlorentinD, line 76: Great, that users of the simple operations will simply benefit from the changes :+1:

main/ejml-simple/src/org/ejml/simple/ops/SimpleOperations_DSCC.java
```diff
@@ -40,6 +43,10 @@
     public transient IGrowArray gw = new IGrowArray();
     public transient DGrowArray gx = new DGrowArray();
 
+    // Workspace for concurrent algorithms
+    public transient GrowArray<Workspace_MT_DSCC> workspaceMT = new GrowArray<>(Workspace_MT_DSCC::new);
+    public transient GrowArray<DGrowArray> workspaceA = new GrowArray<>(DGrowArray::new);
```
FlorentinD, line 48: What is the advantage of the `transient` keyword here. I read, its only useful for serialization.

Also do we assume users of SimpleOperations_DSCC wont several operations in parallel?
Parallel calls would be problematic when the same workspace is used

FlorentinD (approved):
Looks good to me.

No blocking remarks, just some minor comments.
Never heard of the `transient` keyword before this PR

main/ejml-core/src/org/ejml/concurrency/EjmlConcurrency.java
lessthanoptimal: Educated guess. I know for very small matrices it will be a big performance hit so this has to be added. It will definitely be operator dependent. I figured it wasn't worth delaying the feature for a more rigorous performance study that might not happen any time soon. Also people who use SimpleMatrix tend to not be super concerned about small hits to performance. Yes you are right there should be a study done to select this value.

main/ejml-core/src/org/ejml/concurrency/EjmlConcurrency.java
lessthanoptimal: reasonable nit. Coding standards at my previous company. It should also be >= I think too.

main/ejml-simple/src/org/ejml/simple/ops/SimpleOperations_DSCC.java
lessthanoptimal: To be honest I can't remember the exact reason it was used here since that code was written so long ago and there was no comment explaining why.

I had the same thought about parallel operation, but it already had workspace variables that would screw it up. No one has complained yet, but this should be thought through more carefully... Basically there will be no regression but the design might have already been flawed.

lessthanoptimal: just double checked and each SimpleMatrix has its own instance. Not very memory efficient but there will be no threading issue.

PR #148, opened 2021-11-22
Fix #147 and add QOL changes to some Simple classes

### LINKED ISSUE #147: Strange omission of max/minElement from SimpleBase
Hi all,

http://ejml.org/javadoc/org/ejml/simple/SimpleBase.html

SimpleBase has absolute max/minElement but not the normal type- is there a reason behind this omission?

Thanks,
Steve

BluSteve:
The first commit created 3 new files for some reason, although they were just changed by me. The only ambitious change in this PR is probably the column vector constructors, which take in one double/float[] and assume the client wants a [].length by 1 column vector.

### DIFF
--- a/main/ejml-simple/test/org/ejml/simple/TestSimpleMatrix.java
+++ b/main/ejml-simple/test/org/ejml/simple/TestSimpleMatrix.java
@@ -69,6 +69,15 @@ public class TestSimpleMatrix extends EjmlStandardJUnit {
     }
 
     @Test
+    public void constructor_1d_array_simple() {
+        double[] d = new double[]{2, 5, 3, 9, -2, 6, 7, 4};
+        SimpleMatrix s = new SimpleMatrix(d);
+        DMatrixRMaj m = new DMatrixRMaj(8, 1, true, d);
+
+        EjmlUnitTests.assertEquals((DMatrixRMaj)m, (DMatrixRMaj)s.getMatrix(), UtilEjml.TEST_F64);
+    }
+
+    @Test
     public void constructor_2d_array() {
         double[][] d = new double[][]{{1, 2}, {3, 4}, {5, 6}};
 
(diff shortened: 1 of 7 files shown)


lessthanoptimal:
Could you remove those files it created? Files which have float data types are auto generated and that's how it got added. I should create check to make sure that doesn't happen.

I'll take a closer look at the PR soon.

BluSteve:
IIRC you mentioned somewhere that small SimpleMatrix operations are slow because of the need to initialize an empty matrix to dump the results into. Would in-place methods alleviate this issue?

Here's the jblas DoubleMatrix API which I'm drawing a lot of inspiration from.
http://jblas.org/javadoc/index.html

lessthanoptimal:
Looks good but the new functions need unit tests.

lessthanoptimal:
> IIRC you mentioned somewhere that small SimpleMatrix operations are slow because of the need to initialize an empty matrix to dump the results into. Would in-place methods alleviate this issue?

There are a LOT of functions that would need to be added to support in-place operations and I think it would make it a bit too complex. Let's hold off on that for now since people can just use the procedural API if needed. Memory copying with SimpleMatrix only dominates when working with fairly small matrices. 

At one point there was code which experimented with this exact idea. It was simply commented out for a while. What I didn't like about it then was that it broke the nice flow style api.

BluSteve:
It doesn't actually need to break the flow style API as the in place methods can just return the modified object after it's done, which still allows for method chaining.

I actually coded it for personal use already, the number of functions for which this feature applies is surprisingly small (around 10) as both the source and target matrices have to be the exact same size. Otherwise reshaping is afaik just creating a new matrix.

In the JBlas documentation you can see how a similar thing was implemented. e.g. add vs addi, which take in the same parameters and return the same type.

One could argue that it makes the API a bit bloated, but there is a consistent naming scheme (methodName + "i") so it's all in all quite intuitive.

Anyways, I shall finish the unit tests first.

lessthanoptimal:
ok let's get this PR finished then create a new one with the in-place operations since you already coded it up.

BluSteve:
```
public static SimpleMatrix loadBinary( String fileName )
            throws IOException {
        DMatrix mat = MatrixIO.loadBin(fileName);

        // see if its a DMatrixRMaj
        if (mat instanceof DMatrixRMaj) {
            return SimpleMatrix.wrap((DMatrixRMaj)mat);
        } else {
            // if not convert it into one and wrap it
            return SimpleMatrix.wrap(new DMatrixRMaj(mat));
        }
    }
```
By the way, what is your policy on static methods returning SimpleMatrix? I found this method in SimpleBase, and it's the only method in the whole class that returns a SimpleMatrix. All of the other static factory methods are in SimpleMatrix itself, and that's where I put the new ones as well.

lessthanoptimal:
That static method in SimpleBase is probably an oversite from refactoring. Doesn't make sense to return a child object.

lessthanoptimal:
Can you squash the commits? Otherwise this looks good.

main/ejml-simple/test/org/ejml/simple/TestSimpleMatrix.java
```diff
@@ -496,6 +525,18 @@ public void getIndex() {
         assertEquals(((DMatrixRMaj)a.mat).getIndex(0, 2), a.getIndex(0, 2), 1e-6);
     }
 
+    @Test
+    public void toArray2() {
+        SimpleMatrix a = SimpleMatrix.random_DDRM(3, 2, 0, 1, rand);
+
+        double[][] array2 = a.toArray2();
```
lessthanoptimal, line 532: Could you add a check to make sure the shapes match?

lessthanoptimal (changes requested):
One minor nit. Otherwise LGTM. Squash and I'll approve the PR. Thanks!

BluSteve:
Changes made and commits squashed. Thank you!

lessthanoptimal (approved):

PR #149, opened 2021-11-25
Add in-place methods to SimpleBase

BluSteve:
One thing worth noting is that the CommonOps transpose method apparently does not work in-place.

### DIFF
--- a/main/ejml-simple/src/org/ejml/simple/SimpleBase.java
+++ b/main/ejml-simple/src/org/ejml/simple/SimpleBase.java
@@ -154,6 +154,12 @@ public abstract class SimpleBase<T extends SimpleBase<T>> implements Serializabl
         return ret;
     }
 
+    public T transposei() {
+        ops.transpose(mat, mat);
+
+        return (T) this;
+    }
+
     /**
      * <p>
      * Returns a matrix which is the result of matrix multiplication:<br>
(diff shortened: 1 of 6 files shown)


FlorentinD:
Hey @BluSteve,
Is there a reason why you need to use the `Simple` interface?
If you want more control, you could directly use the `CommonOps_DDRM` methods instead of the simple interface.

Personally, I don't see the advantage of putting all these in-place methods on the Simple interface.
I would even argue, this makes it harder for beginners as its yet another concept they need to think of.

FlorentinD:
I just saw you already discussed my concern in the previous PR.

BluSteve:
Honestly it does have a pretty niche use case, and I would understand if the PR was rejected.

The main use would be as a direct find-and-replace for the pattern x=x+a. In my experience, scalar operations benefit most because they aren't slow enough for the memory allocation bottleneck to disappear and they're also intuitive since it's just x+=a.

I personally don't think it's that much of a mental overhead, but then again I am used to JBlas syntax.

lessthanoptimal:
Mentioned this in the other thread, but basically I think now is a good time reconsidering how simple SimpleMatrix is. Most similar libraries have a lot more functions and it is easier to find operations when everything is derived from a single class. Although SimpleMatrix was very much a reaction against that paradigm. Maybe the way to handle this is to have @BluSteve experiment with new designs in the "experimental" package that's not included by default. Then we can consider moving some new operations from there into SimpleMatrix at a later date. That way we don't need to guess what's best right now and can have a less strict PR process.

lessthanoptimal:
Any random people looking at this PR have any thoughts on adding a lots of new functions for in place operations? Main argument against is all the complexity it would add and most functions in SimpleMatrix return a new instance so people would need to be careful.

PR #154, opened 2022-06-29
Fix docstrings for SimpleBase and SimpleMatrix

JozsefKutas:
Various minor docstring fixes for `SimpleBase` and `SimpleMatrix`:
- fixed `SimpleBase#cols` to refer to columns rather than rows
- deleted paragraph referring to variable arguments which are not actually variable
- added missing docstrings for `SimpleMatrix` float constructors
- corrected a few typos

### DIFF
--- a/main/ejml-simple/src/org/ejml/simple/SimpleMatrix.java
+++ b/main/ejml-simple/src/org/ejml/simple/SimpleMatrix.java
@@ -97,15 +97,6 @@ public class SimpleMatrix extends SimpleBase<SimpleMatrix> {
      * column-major.
      * </p>
      *
-     * <p>
-     * Note that 'data' is a variable argument type, so either 1D arrays or a set of numbers can be
-     * passed in:<br>
-     * SimpleMatrix a = new SimpleMatrix(2,2,true,new double[]{1,2,3,4});<br>
-     * SimpleMatrix b = new SimpleMatrix(2,2,true,1,2,3,4);<br>
-     * <br>
-     * Both are equivalent.
-     * </p>
-     *
      * @param numRows The number of rows.
      * @param numCols The number of columns.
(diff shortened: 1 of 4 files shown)


JozsefKutas:
I'm not actually sure why the `SimpleMatrix​(int, int, boolean, double[])` constructor doesn't take varargs, while the `DMatrixRMaj(int, int, boolean, double...)` constructor does, but for the moment I just changed the docstring so that it would be less confusing.

main/ejml-simple/src/org/ejml/simple/SimpleMatrix.java
```diff
@@ -97,15 +97,6 @@ public class SimpleMatrix extends SimpleBase<SimpleMatrix> {
      * column-major.
      * </p>
      *
```
lessthanoptimal: Maybe just change it to vararg? I'm honestly not sure why it isn't either...

lessthanoptimal:
Looks good and thanks for the edits. Just that one comment on the vararg issue you pointed out.

main/ejml-simple/src/org/ejml/simple/SimpleMatrix.java
JozsefKutas: OK, I've changed the constructors to accepts varargs and readded the explanation in the docstring.

lessthanoptimal: can you squash these changes? I'll then approve it.

JozsefKutas: I've squashed the docstring changes, but kept the varargs change separate as it's the only change that affects the code. I can squash that too, if you prefer.

lessthanoptimal (approved):
LGTM

PR #157, opened 2022-09-16
Fix SimpleMatrix type conversion

### LINKED ISSUE #156: SimpleMatrix concatenation ignores type of first matrix when resolving result type
Hi,

I came across a bug when concatenating SimpleMatrix instances, where it seems the type of the first matrix is ignored:

```java
SimpleMatrix fscc = new SimpleMatrix(1, 1, MatrixType.FSCC);
SimpleMatrix ddrm = new SimpleMatrix(1, 1, MatrixType.DDRM);
System.out.println(fscc.concatColumns(ddrm).getType());  // this is DDRM
System.out.println(ddrm.concatColumns(fscc).getType());  // this is FSCC
System.out.println(ddrm.concatColumns().getType());  // also FSCC
```

The issue seems to be in the `specify0( SimpleBase a , SimpleBase ...inputs )` method of the `AutomaticSimpleMatrixConvert` class, which returns `specify(inputs)` when it should probably return `specify(array)`.

JozsefKutas:
Fixes #156.

I added a some extra assertions to existing tests - the extra assertions for `concatRows` and `concatColumns` fail without the fix.

### DIFF
--- a/main/ejml-simple/src/org/ejml/simple/AutomaticSimpleMatrixConvert.java
+++ b/main/ejml-simple/src/org/ejml/simple/AutomaticSimpleMatrixConvert.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2020, Peter Abeles. All Rights Reserved.
+ * Copyright (c) 2022, Peter Abeles. All Rights Reserved.
  *
  * This file is part of Efficient Java Matrix Library (EJML).
  *
@@ -15,7 +15,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.ejml.simple;
 
 import org.ejml.data.Matrix;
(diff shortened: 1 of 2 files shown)


main/ejml-simple/src/org/ejml/simple/AutomaticSimpleMatrixConvert.java
```diff
@@ -35,7 +34,7 @@ public void specify0( SimpleBase a , SimpleBase ...inputs ) {
         SimpleBase array[] = new SimpleBase[inputs.length+1];
         System.arraycopy(inputs,0,array,0,inputs.length);
         array[inputs.length] = a;
-        specify(inputs);
+        specify(array);
```
lessthanoptimal, line 37: well that was a silly type-o. Did all the work then passed in the wrong object. Thanks for the fix!

lessthanoptimal (approved):
Great PR and thanks for updating the unit tests!

lessthanoptimal:
Might do a minor release just for this fix since it could be a while before there's a major new release. Thanks!

Issue #155, opened 2022-06-29
Simplify getting and setting SimpleBase rows/columns


JozsefKutas:
I've recently started using EJML, and the `SimpleBase` API seems a bit clunky when it comes to getting/setting rows/columns.

1. The `SimpleBase#extractVector` takes a boolean parameter which determines whether a row or column is extracted. I don't think theres any obvious reason why `false` should correspond to columns and `true` to rows, so it's easy to get the two confused.

2. `SimpleBase#extractVector` returns a matrix, while `SimpleBase#setRow` and `SimpleBase#setColumn` take a double array, so, for example, to set one row/column to another row/column requires something like:

```java
matrix.setColumn(j, 0, matrix.extractVector(false, j).getDDRM().data);
```

vs. Apache Commons:

```java
matrix.setColumn(j, matrix.getColumn(j));
```

To fix 1, maybe `SimpleBase#extractVector` could be supplemented with methods `extractRow` and `extractColumn` (or `row` and `column`, to be consistent with `SimpleBase#rows` and `SimpleBase#columns`). To fix 2, `SimpleBase#setRow` and `SimpleBase#setColumn` could be overloaded to allow for matrix row/column arguments.

Otherwise I'm enjoying using EJML, so thanks!

lessthanoptimal:
Thanks for the suggestions! Yeah I think SimpleMatrix is in need of a refresh. I'll look into these suggestions. Feel free to post other ones too!

lessthanoptimal:
Just letting you know I plan to work on this soon.

JozsefKutas:
OK, thanks for the heads up. I had a few more thoughts after using EJML a bit more, but didn't get around to putting them in writing. I'm a bit bogged down with work atm, but should be able to respond to this issue and the other either tonight or tomorrow.

lessthanoptimal:
I can relate to that.

JozsefKutas:
Hi,

A few thoughts, in no particular order:

1. The `SimpleMatrix` documentation describes it as a wrapper around `DMatrixRMaj` when it can actually wrap other matrix types.

2. Method naming is sometimes a bit inconsistent: some methods use the `get` prefix (`get`, `getNumElements`, ...), some use `extract` (`extractVector`, ...), some use neither (`rows`, `numRows`, ...). This makes it harder to locate methods in a large API.

[...]

PR #175, opened 2023-01-14
Feature/immutable

lessthanoptimal:

- [x] Your code is covered by tests

### DIFF
--- a/main/ejml-simple/src/org/ejml/simple/ConstMatrix.java
+++ b/main/ejml-simple/src/org/ejml/simple/ConstMatrix.java
@@ -26,12 +26,16 @@ import org.ejml.ops.MatrixIO;
 import java.io.IOException;
 
 /**
- * An immutable (read-only) {@link SimpleMatrix}. No options will modify this state of this matrix and all
- * operations which return a matrix return a new instance with the end result.
+ * A "shallow immutable" matrix that implements all the readonly functions in {@link SimpleBase}. None of the
+ * function calls will modify the original matrix and a new matrix instance is always returned. It's "shallow
+ * immutable" because it's possible to downcast into a modifiable data type or to simply save a reference to
+ * the original matrix and modify that. This interface acts as a strong suggestion that the matrix
+ * should not be modified and that one should not modify it externally. However, the only way to ensure that
+ * no external code modifies this matrix is to create a local copy that can't be accessed externally.
  *
  * @author Peter Abeles
  */
-public interface ImmutableMatrix {
+public interface ConstMatrix<T extends ConstMatrix<T>> {
     /**
      * <p>
      * Returns the transpose of this matrix.<br>
@@ -41,13 +45,13 @@ public interface ImmutableMatrix {
      * @return A matrix that is n by m.
      * @see CommonOps_DDRM#transpose(DMatrixRMaj, DMatrixRMaj)
      */
-    <T extends ImmutableMatrix> T transpose();
+    T transpose();
 
     /**
      * Returns a matrix that is the conjugate transpose. If real then this is the
      * same as calling {@link #transpose()}.
      */
-    <T extends ImmutableMatrix> T transposeConjugate();
+    T transposeConjugate();
 
     /**
      * <p>
@@ -62,7 +66,7 @@ public interface ImmutableMatrix {
      * @return The results of this operation.
      * @see CommonOps_DDRM#mult(DMatrix1Row, DMatrix1Row, DMatrix1Row)
      */
-    <T extends ImmutableMatrix> T mult( ImmutableMatrix B );
+    T mult( ConstMatrix<?> B );
 
     /**
      * <p>
@@ -75,7 +79,7 @@ public interface ImmutableMatrix {
      * @return Kronecker product between this matrix and B.
      * @see CommonOps_DDRM#kron(DMatrixRMaj, DMatrixRMaj, DMatrixRMaj)
      */
-    <T extends ImmutableMatrix> T kron( ImmutableMatrix B );
+    T kron( ConstMatrix<?> B );
 
     /**
      * <p>
@@ -89,7 +93,7 @@ public interface ImmutableMatrix {
      * @param B m by n matrix. Not modified.
      * @return The results of this operation.
      */
-    <T extends ImmutableMatrix> T plus( ImmutableMatrix B );
+    T plus( ConstMatrix<?> B );
 
 
     /**
@@ -105,7 +109,7 @@ public interface ImmutableMatrix {
      * @return The results of this operation.
      * @see CommonOps_DDRM#subtract(DMatrixD1, DMatrixD1, DMatrixD1)
      */
-    <T extends ImmutableMatrix> T minus( ImmutableMatrix B );
+    T minus( ConstMatrix<?> B );
 
     /**
      * <p>
@@ -122,7 +126,7 @@ public interface ImmutableMatrix {
      * @return The results of this operation.
      * @see CommonOps_DDRM#subtract(DMatrixD1, double, DMatrixD1)
      */
-    <T extends ImmutableMatrix> T minus( double b );
+    T minus( double b );
 
     /**
      * Subtracts a complex scalar from each element in the matrix. If the matrix is real, then it will
@@ -132,7 +136,7 @@ public interface ImmutableMatrix {
      * @param imag Imaginary component of scalar value
      * @return The results of this operation.
      */
-    <T extends ImmutableMatrix> T minusComplex( double real, double imag );
+    T minusComplex( double real, double imag );
 
 
     /**
@@ -150,7 +154,7 @@ public interface ImmutableMatrix {
      * @return A matrix that contains the results.
      * @see CommonOps_DDRM#add(DMatrixD1, double, DMatrixD1)
      */
-    <T extends ImmutableMatrix> T plus( double b );
+    T plus( double b );
 
     /**
      * Adds a complex scalar from each element in the matrix. If the matrix is real, then it will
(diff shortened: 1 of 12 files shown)


lessthanoptimal:
Previous requests for an ImmutableMatrix https://github.com/lessthanoptimal/ejml/issues/155 . Ok I can find only one, but I know this has come up several times previously.

lessthanoptimal:
@JozsefKutas @ennerf @FlorentinD Does this look good? The PR adds an ImmutableMatrix which is an interface SimpleMatrix implements.

lessthanoptimal:
Might rename it from ImmutableMatrix to ConstMatrix. Const is shorter...

JozsefKutas:
Hi, I've just looked over the changes and I have a few thoughts.

At the moment, the `ImmutableMatrix` isn't actually immutable. A class is immutable if it comes with a gaurantee that instances can't be modified. There's a quick summary of what this implies at https://math.oxford.emory.edu/site/cs171/immutability/ (taken I think from Bloch's Effective Java).

With the current implementation, there's no guarantee that the underlying data of `ImmutableMatrix` will not be modified after the immutable matrix is created:

```java
DMatrixRMaj matrix = CommonOps_DDRM.identity(2);
ImmutableMatrix immutable = SimpleMatrix.wrap(matrix);
matrix.set(1, 1, 0d);
```

It's closer to the unmodifiable collections returned by some methods in `java.util`. Unmodifiability is a weaker condition than immutablility, but is still useful (see https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/Collection.html#unmodifiable).

It's not strictly unmodifiable either, because it's possible to mutate elements by downcasting:

```java
ImmutableMatrix immutable = SimpleMatrix.identity(2);
SimpleMatrix simple = (SimpleMatrix) immutable;
simple.set(1, 1, 0d);
```

If you go with this implementation then I'd change the name, just to make it clearer to the user what guarantees the class does and doesn't provide.

Alternative approaches would be:

- Have `ImmutableMatrix` extend `SimpleBase`, and throw an `UnsupportedOperationException` in public methods that mutate the underlying data. Make sure methods like `wrap` make copies. If any public mutator methods in `SimpleBase` are used internally in the implementation of non-mutator methods in `SimpleBase` (to prepare the returned matrix) then this will require workarounds. This is the approach taken by Guava (https://github.com/google/guava/wiki/ImmutableCollectionsExplained).

- Rename the current `ImmutableMatrix` interface, and use it as the parent for mutable matrices (`SimpleMatrix`) and immutable matrices. This is similar to the approach used in scala (https://docs.scala-lang.org/overviews/collections-2.13/overview.html). This approach is more complicated, but it avoids using runtime errors to enforce immutability.

Besides these issues, I don't understand the motiviation behind removing the type parameter from `SimpleBase`. It means errors like this aren't caught until runtime:

```java
ImmutableMatrix matrix = SimpleMatrix.identity(2);
StatisticsMatrix scaled = matrix.scale(2d);
// Exception in thread "main" java.lang.ClassCastException: class org.ejml.simple.SimpleMatrix cannot be cast to class org.ejml.example.StatisticsMatrix ...
```

FlorentinD:
In general I like the idea of having an immutable matrix. One use-case could be for input data which you want to make sure you dont modify by accident.

For the PR itself, I agree with Jozsef and I am also missing some tests which show the Exception is thrown.

Also, starting with the API over the SimpleMatrix makes sense, but I could also see this as useful for more concrete implementations as this is a feature more advanced users would use.

lessthanoptimal:
> Hi, I've just looked over the changes and I have a few thoughts.
> 
> At the moment, the `ImmutableMatrix` isn't actually immutable. A class is immutable if it comes with a gaurantee that instances can't be modified. There's a quick summary of what this implies at https://math.oxford.emory.edu/site/cs171/immutability/ (taken I think from Bloch's Effective Java).

Ok Immutable isn't the best description clearly. It's main design goal is to make it so that someone has to go out of their way to modify it. Kinda of like const in C++, which apparently is considered "a shallow immutable object" as you can strip it of the const property in various ways.

Trying to prevent someone from downcasting and modifying or any other "attack" I think is overkill. This is mainly to keep honest people honest and make it easy to see if someone is deliberately breaking the contract. What I don't like about Unmodifiable is that some of the checks are at runtime, which can result in a nasty surprise.

Maybe call it ConstMatrix?

> * Rename the current `ImmutableMatrix` interface, and use it as the parent for mutable matrices (`SimpleMatrix`) and immutable matrices. This is similar to the approach used in scala (https://docs.scala-lang.org/overviews/collections-2.13/overview.html). This approach is more complicated, but it avoids using runtime errors to enforce immutability.

That's what currently done I believe.

> Besides these issues, I don't understand the motiviation behind removing the type parameter from `SimpleBase`. It means errors like this aren't caught until runtime:

Just reverted that change. It was result of an earlier experiment that didn't go anywhere and I forgot to undo it.

Thanks for all the thoughtful comments!

@JozsefKutas

lessthanoptimal:
See [KalmanFilterSimple](https://github.com/lessthanoptimal/ejml/blob/feature/immutable/examples/src/org/ejml/example/KalmanFilterSimple.java) for an example of how this could be put into practice.

lessthanoptimal:
Unless there's some additional feedback or suggestion for a better name, I'm going to merge this in soon. Planning on doing another release soon too.

main/ejml-simple/src/org/ejml/simple/SimpleBase.java
```diff
@@ -241,20 +210,9 @@ public T kron( T B ) {
...
-     * where c is the returned matrix, a is this matrix, and b is the passed in matrix.
-     * </p>
-     *
-     * @param B m by n matrix. Not modified.
-     * @return The results of this operation.
-     * @see CommonOps_DDRM#mult(DMatrix1Row, DMatrix1Row, DMatrix1Row)
-     */
-    public T plus( T B ) {
+    /** {@inheritDoc} */
+    @Override public T plus( ConstMatrix _B ) {
```
JozsefKutas:
```suggestion
    @Override public T plus( ConstMatrix<?> _B ) {
```

main/ejml-simple/src/org/ejml/simple/SimpleBase.java
```diff
@@ -373,22 +278,9 @@ public T plusComplex( double real, double imag ) {
...
-     *
-     * <p>NOTE: If the matrix is complex then 'b' will be treated like a complex number with imaginary = 0.</p>
-     *
-     * @param B m by n matrix. Not modified.
-     * @return A matrix that contains the results.
-     * @see CommonOps_DDRM#add(DMatrixD1, double, DMatrixD1, DMatrixD1)
-     */
-    public T plus( double beta, T B ) {
+    /** {@inheritDoc} */
+    @Override public T plus( double beta, ConstMatrix _B ) {
```
JozsefKutas:
```suggestion
    @Override public T plus( double beta, ConstMatrix<?> _B ) {
```

main/ejml-simple/src/org/ejml/simple/SimpleBase.java
```diff
@@ -398,13 +290,9 @@ public T plus( double beta, T B ) {
...
 
-    /**
-     * Computes the dot product (a.k.a. inner product) between this vector and vector 'v'.
-     *
-     * @param v The second vector in the dot product. Not modified.
-     * @return dot product
-     */
-    public double dot( T v ) {
+    /** {@inheritDoc} */
+    @Override public double dot( ConstMatrix _v ) {
```
JozsefKutas:
```suggestion
    @Override public double dot( ConstMatrix<?> _v ) {
```

main/ejml-simple/src/org/ejml/simple/ConstMatrix.java
```diff
@@ -0,0 +1,798 @@
...
+    /**
+     * <p>
+     * Returns a matrix which is the result of matrix multiplication:<br>
+     * <br>
+     * c = a * b <br>
+     * <br>
+     * where c is the returned matrix, a is this matrix, and b is the passed in matrix.
+     * </p>
+     *
+     * @param B A matrix that is n by bn. Not modified.
```
JozsefKutas:
```suggestion
     * @param B A matrix that is n by p. Not modified.
```

JozsefKutas (approved):
I'm fine with the name `ConstMatrix`.

A few thoughts:

- The methods `concatRows`, `concatColumns`, `setRow` and `setColumn` should probably take `ConstMatrix` objects as arguments.

- This PR makes it possible to perform operations with different subclasses of `SimpleBase`. This could be useful in some situations, and I don't see any disadvantages. For example, the following code previously wouldn't compile, but now compiles and runs without problem:

```java
DMatrixRMaj matrix = CommonOps_DDRM.identity(2);
SimpleMatrix a = SimpleMatrix.wrap(matrix);
StatisticsMatrix b = StatisticsMatrix.wrap(matrix);
System.out.println(a.plus(b));
```

- Most methods of `SimpleBase` that accept `ConstMatrix` won't work if the user passes an object implementing `ConstMatrix` that doesn't extend `SimpleBase`. You could prevent this by making `ConstMatrix` a sealed interface and making `SimpleBase` a non-sealed class.

I don't think `ConstMatrix` works for my own use case and I'm not completely sure it's worth the additional API complexity, but I don't see any other issues.

lessthanoptimal:
@JozsefKutas Do you know if sealed types are syntax sugar and generate bytecode compatible with Java 11? EJML compiles Java 17 but creates bytecode for Java 11. EJML will be stuck with Java 11 bytecode for the foreseeable future. Took forever to give up on Java 8.

lessthanoptimal:
As a side note, I'm not really the target audience for ConstMatrix either. I mostly work with the low level API. I think it applies to people who attempt to restrict access as much as possible and feel the need to use final ever place possible. I also liked the Equations API but I think I'm the only user of that API after all these years. That is probably a good candidate to be removed in the future, at least if it wouldn't break a bunch of my own code.

JozsefKutas:
> Do you know if sealed types are syntax sugar and generate bytecode compatible with Java 11?

They're not just syntax sugar, so I'd guess they're not compatible.

Issue #180, opened 2023-03-01
innerProduct and innerProductSelfSymmetrical operation optimized for sparse vectors


spensonshih:
Hi,

This is a proposal to add `innerProduct` /  `innerProductSelfSymmetrical` operation optimised for sparse vectors.

The new version of `innerProduct` will be specialised to take in sparse vectors for `A` and `C`, and a matrix `B`.

scalar = A<sup>T</sup> * B * C

This is a common operation in portfolio risk modelling for computing variance or variance contributions based on a factor risk model. `A` and `C` represent exposures of a security or portfolio to risk factors, and `B` represents the symmetric and positive semi-definite covariance matrix of the risk factors. `B` is dense because factors are related, but `A` and `C` are sparse because individual securities or portfolios often only are exposed to a small fraction of the factors in the model, depending on their asset type. For example, stocks are not directly exposed to interest rates, while bonds are not directly exposed to the stock market (only indirectly). 

Shrinking the matrix is one option, but it is often not viable as the same matrix will be applied repeatedly to millions of securities and portfolios with varying sets of factors.

The current version of innerProduct iterates over the non-zero elements of the matrix, `B`, which is dense here and causes it be an expensive operation.  By providing a version of innerProduct that iterates over the non-zero elements of `A` and `C` instead, this would avoid a lot of unnecessary operations and have a significant performance improvement.

Here's what the implementation might look like:

```java
public static double innerProduct(DMatrixSparseCSC a, DMatrixD1 b, DMatrixSparseCSC c) { 
    double output = 0.0; 
    for (Iterator<DMatrixSparse.CoordinateRealValue> it1 = a.createCoordinateIterator(); it1.hasNext(); ) { 
        DMatrixSparse.CoordinateRealValue c1 = it1.next(); 
 
        double innerOutput = 0.0; 
        for (Iterator<DMatrixSparse.CoordinateRealValue> it2 = c.createCoordinateIterator(); it2.hasNext(); ) { 
            DMatrixSparse.CoordinateRealValue c2 = it2.next(); 
            innerOutput += c2.value * b.get(c1.row, c2.row); 
        } 
        output += c1.value * innerOutput; 
    } 
    return output; 
} 
```

Of course, we’d add checks for the input’s lengths.   

For `innerProductSelfSymmetrical`, this will be specialised to take in sparse vector `A`, and a positive semi-definite covariance matrix `B`. 

scalar = A<sup>T</sup> * B * A

[...]

PR #181, opened 2023-03-08
Add innerProduct for sparse vectors.

spensonshih:
PR for #180 

The new version of `innerProduct` will be specialised to take in sparse vectors for A and C, and a matrix B.

scalar = A<sup>T</sup>^ * B * C

By looping through only non-zero elements of the vectors, avoids a lot of unnecessary operations and have a significant performance improvement when used with sparse vectors.

And `innerProductSelfSymmetrical`, specialised to take in sparse vector A, and a positive semi-definite covariance matrix B.

scalar = A<sup>T</sup> * B * A

Further enhances the innerProduct, utilising the matrix symmetry, it will calculate the product using just the lower (or upper) triangle, multiply this by 2, then add to the product of the matrix diagonal.

- [X] Your code is covered by tests
- [X] You ran `./gradlew spotlessJavaApply`

### DIFF
--- a/main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
+++ b/main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, Peter Abeles. All Rights Reserved.
+ * Copyright (c) 2023, Peter Abeles. All Rights Reserved.
  *
  * This file is part of Efficient Java Matrix Library (EJML).
  *
@@ -15,14 +15,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.ejml.dense.row.mult;
 
+import java.util.Iterator;
+
 import org.ejml.MatrixDimensionException;
+import org.ejml.UtilEjml;
 import org.ejml.data.DMatrix1Row;
 import org.ejml.data.DMatrixD1;
 import org.ejml.data.DMatrixRMaj;
+import org.ejml.data.DMatrixSparse;
+import org.ejml.data.DMatrixSparseCSC;
 import org.ejml.dense.row.CommonOps_DDRM;
+import org.ejml.dense.row.MatrixFeatures_DDRM;
 
 /**
  * <p>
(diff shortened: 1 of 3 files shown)


lessthanoptimal:
thanks for the PR. Will probably end up shuffling around its location the another class, but that can be done later on. 
- A good unit test would be to compare it compare it against two equivalent multiplications.
- A JMH benchmark. There's room for a few optimizations, but it would be interesting to have this implementation as a baseline. Plus a JMH benchmark will be created anyways for regression testing.

Inside of main/ejml-dsparse/src/org/ejml/sparse/csc/mult there are several examples for JMH and you could put it in a new class there.

I think adding a concurrent implementation should be straight forward. That can be done in another PR.

lessthanoptimal:
Ignore comment about unit tests. I missed the ones you added.

spensonshih:
Thanks for the prompt feedback @lessthanoptimal.  I've added the JMH benchmarks.

main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
```diff
@@ -346,4 +348,85 @@ public static double innerProduct( double a[], int offsetA,
...
+
+    /**
+     * scalar = A<sup>T</sup>*B*C
+     *
+     * @param A (Input) vector
+     * @param B (Input) Matrix
+     * @param C (Input) vector
+     */
+    public static double innerProduct(DMatrixSparse A, DMatrix1Row B, DMatrixSparse C) {
+        if (A.getNumElements() < B.numRows)
```
lessthanoptimal: Can you make sure the inputs are vectors and throw an exception if they are not? Also shouldn't it be 'A.getNumElements() != B.numRows'?

If you rebase off of SNAPSHOT there's a new function MatrixFeatures.isVector() that should simplify the code a little bit

main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
```diff
@@ -346,4 +348,85 @@ public static double innerProduct( double a[], int offsetA,
...
+            throw new IllegalArgumentException("Length of 'a' isn't long enough");
+        if (C.getNumElements() < B.numCols)
+            throw new IllegalArgumentException("Length of 'c' isn't long enough");
+
+        VectorEntry[] nzValuesC = nzValues(C);
+        int sizeC = nzValuesC.length;
+
+        double output = 0.0;
+        for (Iterator<DMatrixSparse.CoordinateRealValue> i = A.createCoordinateIterator(); i.hasNext(); ) {
```
lessthanoptimal, line 369: Could you add another version that is specific to `DMatrixSparseCSC`  that directly accesses the non-zero values? I bet that will be a lot faster. It would make sense to keep both versions around.

lessthanoptimal (commented):
Added a couple of requests/comments. Assuming I don't lose power and internet for 3 days again this shouldn't take too long to get approved.

main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
spensonshih: Excellent suggestion.  My benchmarking shows ~10% increased throughput with the direct access to the non-zero values.

main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
```diff
@@ -346,4 +350,153 @@ public static double innerProduct( double a[], int offsetA,
...
+     *
+     * @param A (Input) A vector that has length m.
+     * @param B (Input) A matrix that is m by n.
+     * @param C (Input)  A vector that has length n.
+     */
+    public static double innerProduct(DMatrixSparse A, DMatrix1Row B, DMatrixSparse C) {
+        if (!MatrixFeatures_DDRM.isVector(A))
+            throw new IllegalArgumentException("'A' must be a vector");
+        if (A.getNumElements() != B.numRows)
+            throw new IllegalArgumentException("Length of 'A' isn't long enough");
```
mattsahn: Text of error isn't accurate in some cases if A elements are more than B rows. 

"Length of 'A' vector not equal to number of rows in 'B' matrix" would be more accurate/clear. Same for the other similar error messages throughout.

main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
lessthanoptimal: Was expecting it to be more, but still an improvement!

main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
lessthanoptimal: I agree.

main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
```diff
@@ -346,4 +350,153 @@ public static double innerProduct( double a[], int offsetA,
...
+
+    /**
+     * scalar = A<sup>T</sup>*B*C
+     *
+     * @param A (Input) A vector that has length m.
+     * @param B (Input) A matrix that is m by n.
+     * @param C (Input)  A vector that has length n.
+     */
+    public static double innerProduct( DMatrixSparseCSC A, DMatrixRMaj B, DMatrixSparseCSC C) {
+        if (!MatrixFeatures_DDRM.isVector(A))
```
lessthanoptimal: might as well make this arguments check a function to avoid code duplication

main/ejml-ddense/src/org/ejml/dense/row/mult/MatrixVectorMult_DDRM.java
```diff
@@ -346,4 +350,153 @@ public static double innerProduct( double a[], int offsetA,
...
+        double output = 0.0;
+        for (int i = 0; i < A.nz_length; i++) {
+            int index1 = A.nz_rows[i];
+            double value1 = A.nz_values[i];
+            int b_offset = index1 * B.numCols;
+            double diagonalValue = B.data[b_offset + index1];
+            double sum = 0.0;
+            for (int j = i + 1; j < A.nz_length; j++) {
+                int index2 = A.nz_rows[j];
+                double value2 = A.nz_values[j];
```
lessthanoptimal: nit. I would just inline value2 into the line below

lessthanoptimal:
almost there

lessthanoptimal:
Can you squash it into a single commit? When I have a break from this conference I'll do a local checkout then hopefully approve it. As a warning, after this gets merged I'm going to create a PR which does some refactoring. Mostly move the location of files. It would take longer to explain where everything should go than to do it.

spensonshih:
@lessthanoptimal Apologies for the delay, I had some hardware troubles yesterday.  I've now squashed into a single commit.

lessthanoptimal (approved):
LGTM thanks for these new features!

lessthanoptimal:
looks like a spotless issue snuck in.  `./gradlew :main:spotlessApply` squash again, then I'll merge it.

spensonshih:
Fixed the spotless issue and squashed.

Issue #182, opened 2023-03-14
Bug in MatrixMatrixMult_MT_DDRM.multTransA and concurrent silent failures


hageldave:
Hi, first off, love your work.

Now, I wanted to calculate transpose(A)*B with large `DMatrixRMaj` matrices. This would trigger the multithreaded implementation when using `SimpleOperations_DDRM`. However, something goes wrong and an exception is thrown, but it is caught and the stacktrace is printed.

Consider the following code:
```java
import org.ejml.data.DMatrixRMaj;
import org.ejml.dense.row.CommonOps_MT_DDRM;

public class EJMLThings {

	public static void main(String[] args) {
		int n=1_000_000;
		DMatrixRMaj a = new DMatrixRMaj(n, 4);
		DMatrixRMaj b = new DMatrixRMaj(n, 4);
		DMatrixRMaj c = new DMatrixRMaj(a.numCols, b.numCols);

		// now calculate c=trp(a)*b

		/* Theses two lines will result in calling the multi-threaded 
		 * multTransA implementation when matrix 'a' is large
		 * 
		 *   SimpleOperations_DDRM ops = new SimpleOperations_DDRM();
		 *   ops.multTransA(a, b, c);
		 *   
		 * So effectively calling: CommonOps_MT_DDRM.multTransA(a, b, c);
		 */
		CommonOps_MT_DDRM.multTransA(a, b, c);
	}
}
```

As already said, something goes wrong, and I'll put the output of the execution next, **but that is not the point**.
The point is, that the exception is caught internally. [`ConcurrencyOps.loopFor(...)`](https://github.com/lessthanoptimal/ejml/blob/ed7cad3a76057b3000c3a69d65a21be2488bc76b/main/ejml-core/src/pabeles/concurrency/ConcurrencyOps.java#L61) catches `ExecutionException` which is any exception that happens during the submitted task. Then it prints a stacktrace and the exception vanishes into thin air. 
That is a problem, cause I want my application to crash miserably in such a situation and not assume everything works cause I missed the console output. 
Of course, it would be nice if the code was working, but I think there is no unit test for `CommonOps_MT_DDRM`, and it could be that many more methods using `ConcurrencyOps.loopFor(...)` don't work and have slipped unnoticed since they are silenced.
```
java.util.concurrent.ExecutionException: java.lang.ArrayIndexOutOfBoundsException
[...]

PR #183, opened 2023-03-15
Fix silenced exception in concurrent executions and bugfix for concurrent multTransA_reorder

hageldave:
This PR fixes the issues described in #182.
The discovery of a bug in `MatrixMatrixMult_MT_DDRM.multTransA_reorder(a,b,c)` led to the discovery of dangerous behavior of `ConcurrencyOps.loopFor` methods. These methods catch exceptions that are thown in threadpool-submitted tasks and do not propagate these exceptions further as would be the case for the single-threaded counter part.

This PR fixes the bug, changes behavior of `loopFor` methods to propagate exceptions, and adds a unit test for `SimpleOperations_DDRM.multTransA` (which calls the bug affected method under the hood).

- [x] Your code is covered by tests

### DIFF
--- a/main/ejml-core/src/pabeles/concurrency/ConcurrencyOps.java
+++ b/main/ejml-core/src/pabeles/concurrency/ConcurrencyOps.java
@@ -62,7 +62,7 @@ public class ConcurrencyOps {
         try {
             pool.submit(() -> IntStream.range(start, endExclusive).parallel().forEach(consumer)).get();
         } catch (InterruptedException | ExecutionException e) {
-            e.printStackTrace();
+            throw new RuntimeException(e);
         }
     }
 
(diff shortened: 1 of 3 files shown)


lessthanoptimal (commented):
LGTM

lessthanoptimal:
Thanks for finding this issue and fixing it! Only comment I have is that modifying the generic thread comparison to catch it and have it apply to all other functions would be best instead of having a one off test. That can be a separate PR and care needs to be taken to not slow down unit tests unnecessarily.

lessthanoptimal:
Could you also rebase and fix the lint issue?

hageldave:
I think I fixed the lint issue, did not understand what `spotlessJavaApply` does (still not so sure, but I think it corrects format stuff).

To your other comment about modifying generic thread comparison, I don't know what you mean. Is there a specific unit test that would be a better fit to check the fitness of concurrent implementations?

lessthanoptimal:
Spotless is basically a code formatting tool. Here its done some minimal formatting checks and fixes since it doesn't do exactly what I want.

The test comment was that the test really should go in CheckMultiThreadAgainstSingleThread and be applied to everything. I'm looking into that now.

lessthanoptimal (approved):

PR #188, opened 2023-06-22
Documentation additions and changes for SimpleMatrix

JozsefKutas:
This PR contains various documentation additions/changes for `SimpleMatrix`:

- The main addition is a set of tables listing the methods in `SimpleMatrix`/`SimpleBase` with brief descriptions. Unlike the javadoc summary table, this groups and sorts methods according to functionality. Given that `SimpleMatrix`/`SimpleBase` combined have ~100 methods, it can be difficult to find the method you need when looking through an ungrouped alphabetical list.

- Added missing javadoc for a few methods.

- References to an internal `DMatrixRMaj` have been replaced by references to an internal matrix where appropriate.

- Replaced "a.k.a." by "or" to avoid the javadoc summary cutting off early (it assumes a period and space terminates the first sentence).

PR #189, opened 2023-06-22
Add optional class parameter to ones and filled

JozsefKutas:
This adds overloads of `SimpleMatrix#ones` and `SimpleMatrix#filled` with an optional class parameter. This is roughly consistent with `SimpleMatrix#identity` and `SimpleMatrix#diag`.

This PR follows from #188 - the two overlap, so it didn't seem to make sense to separate them. I can separate this if there are problems with #188.


lessthanoptimal:
Ok got the first PR down. Can you rebase and resolve the conflict?

PR #203, opened 2024-12-02
Add publishing for the uberjar artifact

### LINKED ISSUE #39: Java 9 module support
Hello,
Is it possible to add java 9 module support to ejml. 
For the moment, if I create a module:
module EjmlTest {
	requires ejml.core;
	requires ejml.cdense;
	requires ejml.ddense;
	requires ejml.fdense;
	requires ejml.dsparse;
	requires ejml.experimental;
	requires ejml.simple;
	requires ejml.zdense;
}

with a sample test:
package test;
import org.ejml.data.DMatrix3x3;
public class Test {
	public static void main(String[] args) {
		DMatrix3x3 a = new DMatrix3x3();
		a.print();
	}
}

I get the following error message:
Error occurred during initialization of boot layer
java.lang.module.ResolutionException: Modules ejml.ddense and ejml.experimental export package org.ejml.dense.row.decomposition.lu to module EjmlTest

Thanks!

SamCarlberg:
Resolves #39 by providing a single uberjar with an automatic module name, which modular libraries and applications can safely consume. Modularizing each artifact would require changing package names, which would break existing users.

Publishing uses the `ejml-one` artifact ID using `./gradlew :publish` on the root project, and sets the automatic module name to `ejml.all`.

The uberjar artifact is named `ejml-one`, since `ejml-all` is already used by the depends-on-everything project. I don't know if changing the existing `ejml-all` artifact to publish the uberjar would break users, so I erred on the side of caution here.


PeterJohnson:
Would ejml-unified be a better name?

lessthanoptimal:
wait so each project package doesn't need to have it's own module name? For maven central this has to be a seperate uber jar right?

SamCarlberg:
Each subproject would ideally have its own java module (eg ejml.cdense, ejml.core, etc), but classes in the `org.ejml` package are exported from different subprojects, which the java module system prohibits.

Refactoring all the subprojects to export different subpackages instead of sharing the org.ejml package would break users, so instead this PR creates a fat jar that bundles _everything_ together, and sets an automatic module name for compatibility with the module system. The fat jar would need to be published as a new maven artifact.

lessthanoptimal:
I'm trying to figure out a way avoid code duplication. That means use the existing publish. Here's my attempt to use shadowjar to create the uber jar. The current issue is that the name conflcits with the jar in `components.java` my attempts to override the defaults in `components.java` with the output from shadowjar have been unsuccessful. Worst case I'll fall back to your approach.

https://github.com/lessthanoptimal/ejml/pull/204

lessthanoptimal:
The PR I referenced above is publishing to maven local at least now. No source or javadoc jars yet. Could you or @PeterJohnson  check to see if this works? I should be `ejml-java9module`

SamCarlberg:
@lessthanoptimal I've checked out #204 and left a comment, I think that PR makes more sense as a baseline than this one but wasn't quite right for a modularized usecase. I'll close this PR in favor of yours

PR #204, opened 2025-05-08
feature/java9module

lessthanoptimal:
- WIP for using shadowjar to combine dependencies into an uber jar for publishing
- jar name conflicts with what's in components.java

- [ ] Your code is covered by tests
- [ ] You ran `./gradlew spotlessJavaApply`


SamCarlberg:
@lessthanoptimal From some testing with a local publish, it doesn't look like this will work for a modularized JAR.
- The `-ignore` artifact is pulled in preferentially over the shadowed JAR and includes an automatic module name in the manifest, but doesn't include any class files
- The normal EJML subprojects (ejml-simple, ejml-cdense, et cetera) are pulled in as transitive dependencies by the `-ignore` artifact and appear on the classpath


I had to make these changes to get the published artifact working locally:
- Change dependencies from `implementation` to `compileOnly`, so they wouldn't be transitively required by my library
  - This meant explicitly listing them as entries in the shadowjar, as they would be excluded otherwise
- Removing the `-ignore` artifact by setting the archive classifier to `null` instead of `ignore` or `''`

This resulted in a single JAR being published (`ejml-java9module-0.44.1-SNAPSHOT.jar`) containing all of the EJML classes and a manifest with the automatic module name. The POM doesn't list any dependencies, so there's no risk of the individual subprojects being pulled in transitively and causing problems on the classpath

<details>
<summary>git diff</summary>

```git
diff --git a/main/ejml-java9module/build.gradle b/main/ejml-java9module/build.gradle
index 4f13afa8..b1b5c051 100644
--- a/main/ejml-java9module/build.gradle
+++ b/main/ejml-java9module/build.gradle
@@ -1,25 +1,27 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
 plugins {
     id 'com.github.johnrengelman.shadow' version '8.1.1' // Or the latest version
 }
 
 dependencies {
-    implementation project(':main:ejml-core')
-    implementation project(':main:ejml-fdense')
-    implementation project(':main:ejml-ddense')
-    implementation project(':main:ejml-cdense')
-    implementation project(':main:ejml-zdense')
-    implementation project(':main:ejml-dsparse')
-    implementation project(':main:ejml-simple')
+    compileOnly project(':main:ejml-core')
+    compileOnly project(':main:ejml-fdense')
+    compileOnly project(':main:ejml-ddense')
+    compileOnly project(':main:ejml-cdense')
+    compileOnly project(':main:ejml-zdense')
+    compileOnly project(':main:ejml-dsparse')
+    compileOnly project(':main:ejml-simple')
 }
 
 // Free up the no name classifier for shadowJar
 tasks.named('jar', Jar) {
-    archiveClassifier = 'ignore'
+    archiveClassifier = null
 }
 // NOTE: downside to this approach is that there's now a junk jar named `ignore` being published
 
 shadowJar {
-    archiveClassifier.set('')                  // no “-all” suffix <--- this also causes a conflict
+    archiveClassifier.set(null)                  // no “-all” suffix <--- this also causes a conflict
 
     // Explicitly configure the modules to include
 //    from(sourceSets.main.runtimeClasspath.files) {
@@ -27,10 +29,28 @@ shadowJar {
 //    }
 }
 
+tasks.named('shadowJar', ShadowJar) {
+    configurations = project.configurations.named('compileClasspath').map { [it] }.get()
+    dependencies {
+        include(project(':main:ejml-core'))
+        include(project(':main:ejml-fdense'))
+        include(project(':main:ejml-ddense'))
+        include(project(':main:ejml-cdense'))
+        include(project(':main:ejml-zdense'))
+        include(project(':main:ejml-dsparse'))
+        include(project(':main:ejml-simple'))
+    }
+}
+
+
 //java {
 //    withSourcesJar() // Optional: Include sources in the combined JAR
 //    withJavadocJar() // Optional: Include Javadoc in the combined JAR
 //}
 
-jar { manifest { attributes('Automatic-Module-Name': 'ejml.unified') } }
+jar {
+    manifest {
+        attributes('Automatic-Module-Name': 'ejml.unified')
+    }
+}
```
<details>

SamCarlberg:
@lessthanoptimal Any plans on finalizing this?

lessthanoptimal:
@SamCarlberg I've updated it with your modifications to run in the new build architecture.  EJML just had its entire Gradle build script modernized. Can you verify it works? I'm trying to figure out an automated test to verify it works as intended in the future. I also need to verify that it pushes java9module and need to update the readme with instructions.

lessthanoptimal:
After some heavy lifting from Claude I think it might be ready. Can you try publishing to maven local and follow the instructions in the readme and let me know if everything works?

SamCarlberg:
@lessthanoptimal Thanks! I'll take a look and test it out locally.

SamCarlberg (approved):
Tested in a local project, works fine. Using `requires transitive` on an automatic module causes javac warnings, so an actual `module-info.java` file in the java9module project could be worth considering, but it might be difficult to set up

lessthanoptimal:
@SamCarlberg Just pushed some new code that seems to fix that issue. Can you test again and let me know if it works? There's a force push so you should fetch and do a hard reset.

main/ejml-java9module/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+tasks.named<Jar>("jar") {
+    // Put module-info.class in first so EXCLUDE always prefers ours
+    from(compileModuleInfo)
+
+    // Unzip each module's classes/resources into this jar.
+    from(bundled.elements.map { locations -> locations.map { zipTree(it.asFile) } }) {
+        // Don't let the bundled jars' own manifests/signatures clobber ours.
+        exclude("META-INF/MANIFEST.MF", "META-INF/*.SF", "META-INF/*.RSA", "META-INF/*.DSA")
+    }
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
```
SamCarlberg, line 67: So this generates the uberjar, but the javadoc and sources jars are empty (I should have caught this yesterday, my bad). Can something similar be done to generate the other jars so the publication is usable standalone?

main/ejml-java9module/module/module-info.java
```diff
@@ -0,0 +1,63 @@
+module ejml.java9module {
+    requires transitive java.desktop;
```
SamCarlberg: Does this need to be a transitive requirement? As far as I can tell, java.awt.BufferedImage is only used by a couple of classes for rendering matrices as images in a GUI popup, which doesn't seem like a core usecase. Changing it to a static requirement would make it only needed at compile-time; if downstream consumers need the GUI previews, they'd just need to have the desktop module available.

My usecase is in a headless embedded environment, which may not even have the java.desktop module available. A transitive requirement would make our applications crash at startup when loading the EJML module, while a static requirement would allow the library to load (and subsequently crash at runtime iff someone tries to load the classes that use the desktop APIs)

https://dev.java/learn/modules/optional-dependencies/

SamCarlberg (changes requested):

main/ejml-java9module/module/module-info.java
lessthanoptimal: done

lessthanoptimal:
@SamCarlberg The latest (forced push) update should address those two issues. You will see javadoc as empty for SNAPSHOT builds since its disabled to speed up the build. To test it you should comment out the line below in `ejml.java-concentions.gradle.kts`

```
enabled = !project.version.toString().contains("SNAPSHOT")
```

SamCarlberg (approved):
Looks good! Source and javadoc jars are populated and the module is usable with no issues.

lessthanoptimal:
Thanks for all the help!

